PR #15, opened 2015-02-10
minor changes for University and Person names with schema.org markup

nateprewitt:
Added a couple additional schema.org itemscope changes to some of the propStatement files. I wrapped some text in spans which shouldn't cause css formatting issues unless specifically targeted.


tworrall:
The changes for the fullName template are fine, but I'm not sure we should go with the changes to the educationalTraining template. In most cases the item type "CollegeOrUniversity" applies, but educational training also includes medical residencies, which can be done at non-teaching hospitals, and internships, such as judicial internships.

Nate?

nateprewitt:
Yeah, I think CollegeOrUniversity may have been a bit short sighted. Sounds like your solution in the merge will work.

PR #16, opened 2015-02-11
Start ABox recompute after searchindexer

grahamtriggs:
Moved the reasonor listener to after the SearchIndexer, so that the searchindexer is started first, ensuring that it is initialized before the reasoner, so that the rebuildIndex won't throw a NullPointerException.


j2blake:
Thanks, Graham - it will likely be a couple of days before I can look at these changes.

j2blake:
Accepting this change, and creating the corresponding change in Vitro.

j2blake:
Reverted this change, after further examination of the inter-relationships among the dozen initializers that might be affected. #17

PR #17, opened 2015-02-15
Revert "Start ABox recompute after searchindexer"

j2blake:
Reverts vivo-project/VIVO#16 - while it might be preferable to start the search indexer before the reasoners, it is not clear that the dozen or so other initializers would not be affected by this change.


grahamtriggs:
OK - I wasn't entirely sure about the other dependencies. But the SearchIndexer can generate NullPointerExceptions and start in a paused state at the moment.

I have another pull request that I can open that prevents the NullPointerException.

PR #18, opened 2015-03-30
fixing missing/bad data for Rwanda

nateprewitt:
Fixed bad JSON data for Rwanda in latLongJson.js. Added a geoClass, mapType, and latitude variable to the Rwanda object to prevent crashes during lookup in homePageMaps.js. No associated ticket.


j2blake:
Thanks, Nate!

PR #22, opened 2015-07-31
Remove outdated references to list_view_configuration_guidelines.txt

gneissone:
See https://jira.duraspace.org/browse/VIVO-1044

### DIFF
--- a/productMods/config/listViewConfig-additionalEmail.xml
+++ b/productMods/config/listViewConfig-additionalEmail.xml
@@ -1,36 +1,36 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
 <!-- $This file is distributed under the terms of the license in /doc/license.txt$ -->
 
-<!-- See guidelines in vitro/doc/list_view_configuration_guidelines.txt -->
+<!-- See guidelines at https://wiki.duraspace.org/x/ExUdB -->
 
 <list-view-config>
-    <query-select>    
+    <query-select>
         PREFIX vcard: &lt;http://www.w3.org/2006/vcard/ns#&gt;
-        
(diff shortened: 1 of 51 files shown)


nateprewitt:
It looks like the link used to replace the list_view_configuration_guidelines.txt points back to a deleted post about job positions. The latest version of the file I could find on the Duraspace wiki is from the 1.7 documentation: https://wiki.duraspace.org/x/9K0OB.

I would suggest we might reintroduce some of this documentation back into the codebase /doc/ directory, instead of pointing to somewhat ephemeral wiki pages. This also keeps the documentation versioned with the associated codebase and prevents issues with finding things when wiki server changes are made.

nateprewitt:
Also, not to nitpick Ben, but it would probably be best to revert your branch back to 7fa0c0a and commit without the changes made in pull request #21 to keep requests separated and the commit history clean(er).

gneissone:
I'm not sure how I messed up that link. 
There's actually a version of the file in the current release space of the wiki: https://wiki.duraspace.org/x/eYXVAw. 
Please pretend this never happened.

nateprewitt:
No need to close the ticket, I can show you how to revert and fix the changes with this ticket/branch if you'd like.

gneissone:
Not sure what best practices are, I figured it would be easier/cleaner to start fresh since the other pull request is referencing the same branch. ¯\_(ツ)_/¯

j2blake:
Thanks for catching this outdated reference. As you noted, this information is available in the Wiki.
It is also available as Section 3.9.2 of "vivo/doc/Changing the Appearance of VIVO.pdf" in the release, this is generated from the Wiki as part of the release process.

PR #43, opened 2016-11-03
VIVO-1286 Updating License Info

nateprewitt:
A lot of the license info for VIVO ended up a bit buried after the Maven rework. I figured if we're going to have a notice about where to find the license in most files, that it should be where we say it is. Placing it in the top level directory isn't necessarily the best option, but seems to be the most common place people look.

We also have our dependency licenses buried in the legacy directory, and I'm not sure what the best way is to expose those, or if it's even necessary. Any input on that would be greatly appreciated.


nateprewitt:
I wanted to ping this issue on Github again since I haven't heard anything on JIRA. It still seems like we should put our license in an easily discoverable place and fix the bad documentation. We're also violating the licenses of a few of our upstreams by not including their respective licenses.

I can bring this back up to date if desired, otherwise I'll plan on closing it out at the end of the week.

grahamtriggs:
Hi Nate,

Please leave it open - it's a useful reminder that it hasn't been done yet, and to be taken care of later.

I've been holding off, primarily because it's administrative stuff that can be picked up whilst people are testing, whereas I need to get some feature work done in order to get to a testing phase.

Alternatively, any committer can just move this forward and make changes to the develop branch.

G

nateprewitt:
Alright, I've got everything updated. I'll plan on merging this afternoon unless there's any issues people would like addressed.

I'll get an analogous PR going for Vitro as well.

PR #44, opened 2016-11-08
Support vCard editors in list views

gneissone:
Per [this discussion](https://groups.google.com/d/topic/vivo-tech/EwDXvadDuL8/discussion), makes it possible to model an editor as a vCard instead of a foaf:Person and display them in the list views for publications and people.

### DIFF
--- a/webapp/src/main/webapp/config/listViewConfig-authorInAuthorship.xml
+++ b/webapp/src/main/webapp/config/listViewConfig-authorInAuthorship.xml
@@ -10,75 +10,101 @@
         PREFIX bibo: &lt;http://purl.org/ontology/bibo/&gt;
         PREFIX vitro: &lt;http://vitro.mannlib.cornell.edu/ns/vitro/0.7#&gt;
         PREFIX foaf: &lt;http://xmlns.com/foaf/0.1/&gt;
+        PREFIX vcard:  &lt;http://www.w3.org/2006/vcard/ns#&gt;
+        PREFIX obo: &lt;http://purl.obolibrary.org/obo/&gt;
 
-        SELECT DISTINCT ?subclass
+        SELECT DISTINCT
             ?authorship
             ?infoResource ?infoResourceName
-            ?dateTime
-            ?journal
-            ?volume
-            ?startPage
-            ?endPage
-            ?doi
-            ?pmid
-            ?isbn10
-            ?isbn13
-            ?publisher
-            ?locale
-            ?appearsIn
-            ?partOf
-            ?editor
-            ?hideThis
+            (sample(?irsubclass) as ?subclass)
+            (sample(?irdateTime) as ?dataTime)
+            (sample(?irjournal) as ?journal)
+            (sample(?irvolume) as ?volume)
+            (sample(?irstartPage) as ?startPage)
+            (sample(?irendPage) as ?endPage)
+            (sample(?irdoi) as ?doi)
+            (sample(?irpmid) as ?pmid)
+            (sample(?irisbn10) as ?isbn10)
+            (sample(?irisbn13) as ?isbn13)
+            (sample(?irpublisher) as ?publisher)
+            (sample(?irappearsIn) as ?appearsIn)
+            (sample(?irpartOf) as ?partOf)
+            (sample(?irlocale) as ?locale)
+            (sample(?irhideThis) as ?hideThis)
+            (group_concat(DISTINCT ?editor; separator='; ') as ?editors)
+            (group_concat(DISTINCT ?editorName; separator = '; ') as ?editorNames)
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-informationResourceInEditorship.ftl
```diff
@@ -12,10 +12,18 @@
 <#-- Use a macro to keep variable assignments local; otherwise the values carry over to the
      next statement -->
 <#macro showEditorship statement>
-    <#if statement.person??>
-        <a href="${profileUrl(statement.uri("person"))}" title="${i18n().editor_name}">${statement.personName}</a>
+    <#if statement.editor??>
+        <#if statement.subclass?? && statement.subclass?contains("vcard")>
+            <#if statement.editorName?replace(" ","")?length == statement.editorName?replace(" ","")?last_index_of(",") + 1 >
```
nateprewitt: Would it be possible to get a comment for what this is doing since it's a bit difficult to visually parse as a one-liner. If I'm understanding this correctly, this is trying to solve the additional comma issue I brought up above?

gneissone:
@nateprewitt A lot of the code is repurposed from elsewhere in the VIVO codebase so I'm certainly not opposed to some optimization and it would be good to try and propagate those changes back.

webapp/src/main/webapp/config/listViewConfig-authorInAuthorship.xml
```diff
@@ -10,75 +10,101 @@
...
-            ?pmid
-            ?isbn10
-            ?isbn13
-            ?publisher
-            ?locale
-            ?appearsIn
-            ?partOf
-            ?editor
-            ?hideThis
+            (sample(?irsubclass) as ?subclass)
```
gneissone: Yes, but VIVO will only use the first result anyway without tricks like the group_concat I'm using here. In order to use group_concat the other variables need to be included in the GROUP BY statement or limited to one with SAMPLE. Otherwise a " Non-group key variable in SELECT" error is thrown.

webapp/src/main/webapp/config/listViewConfig-authorInAuthorship.xml
```diff
@@ -10,75 +10,101 @@
...
                 ?editorship a core:Editorship .
-                ?editorship core:relates ?editorObj .
-                ?editorObj rdfs:label ?editor .
+                ?editorship core:relates ?editor .
+                ?editor a vcard:Kind .
+                ?editor vcard:hasName ?vName .
+                ?vName vcard:familyName ?lastName .
+                OPTIONAL { ?vName vcard:givenName ?firstName . }
+                OPTIONAL { ?vName core:middleName ?middleName . }
+                bind ( COALESCE(?firstName, "") As ?firstName1) .
```
gneissone: The `COALESCE` statements were originally there because `bind` fails if one of the variables is unbound. I added the `IF` statement in the `CONCAT` because the extra spaces around editors sans middles name were bugging me and I don't think SPARQL has a trim function. But you're right, the `COALESCE` statements for the first and middle names are unnecessary now. If we add another `IF` statement for the last name we can get rid of `COALESCE` completely. Any idea on performance between the two?

nateprewitt: Do we need an `IF` for `?lastName`? It's already required to be bound with `?vName vcard:familyName ?lastName .` so we shouldn't be hitting this OPTIONAL block at all if it can't resolve that. Perhaps there's an edge case I'm missing in my tests though.

gneissone: Ah yeah, I see. The last name could be optional, too, to account for Cher and Madonna. Then again, it's likely required in a handful of other list view configs.

gneissone:
Since this came up in the development call, it would be good to know if there is still interest in this PR. Aside from the comments above, an additional concern was raised that VIVO may change the ontology so the vCard-style of modeling external people would be abandoned. 

That said, this code is being used in production on Connect UNAVCO since we needed to display external PIs and Co-PIs without linking to empty vCard pages. [Example here](https://connect.unavco.org/display/awd724371).

mconlon17:
No plans to move away from Vcard.  Vcard will be updated in Sprint 1.

gneissone:
This pull request has been stagnant for far too long. I still think this is a worthy improvement for anyone that utilizes the UI to edit the database (given the current VIVO schema) but given that the preferred list view format has changed, necessitating changes to this PR, and that I haven't heard any demands for this change in the interim, I will close the pull request.

PR #46, opened 2016-12-12
[VIVO-1063] Update jQuery to 1.12.4 and associated upgrades

grahamtriggs:


webapp/src/main/webapp/js/individual/individualQtipBubble.js
```diff
@@ -6,48 +6,44 @@ $(document).ready(function(){
...
-                lineHeight: '15px',
                 width: 180,
-                border: 'none',
-                background:'url(' + imagesPath + '/individual/researchAreaBubble.png)  no-repeat'
             }
         });
     });
 
+    $('head').append('<style id="bubbleCSS">.qtip { font-size: 14px; max-width: none !important; } .bubbleTip { url(' + imagesPath + '/individual/toolTipBubble.png) no-repeat; } </style>');
+//    padding: '0em',
```
nateprewitt: Probably can ditch these too.

webapp/src/main/webapp/js/individual/individualQtipBubble.js
```diff
@@ -6,48 +6,44 @@ $(document).ready(function(){
     
     $.extend(this, i18nStrings);
 
+    $('head').append('<style id="researchAreaCSS">.qtip { font-size: 14px; max-width: none !important; } .researchAreaTip { background: url(' + imagesPath + '/individual/researchAreaBubble.png) no-repeat; } </style>');
+    padding: '0em',
```
grahamtriggs: I'm not sure it "should". But in the short term, it needed to be in order to replicate the embedded stylesheet code that was used with qTip, but isn't supported in qTip 2

nateprewitt: `padding: '0em',` in particular? I don't believe that's a valid standalone javascript statement and it's not contained inside of any qTip or function declaration. It looks like it either needs to be put in the qtip styling for `researchAreaCSS` or moved down to the styling for the callback provided to `$('#researchAreaIcon').each`.

grahamtriggs: Ahh... that line should have been commented out. Just there as a placeholder for migrating the CSS from the previous definition. Thanks! I've removed it.

PR #49, opened 2017-02-02
Fix Altmetric badge display for publications with only PMID

smmckay:
When you have an artifact with a PMID but no DOI, the Altmetric badge isn't displayed. Technically there are two changes in one commit here, but they're both one-liners.


grahamtriggs:
Thanks for this!

Pull requests need to be accompanied by a Jira ticket. I have created one for you in this instance:

https://jira.duraspace.org/browse/VIVO-1314

PR #50, opened 2017-02-09
Per VIVO-1316, moving base lookup services to Vitro

hudajkhan:
https://jira.duraspace.org/browse/VIVO-1316


hudajkhan:
To work, this will depend on https://github.com/vivo-project/Vitro/pull/55

PR #53, opened 2017-05-16
VIVO-1319 Remove dependency on sourceforge.net JSON parser.

j2blake:
Created JacksonUtils to help ease the transition. Replace the functionality of JSONSerializer and JSONUtils with plug-compatible methods.
There are no constructors for ArrayNode and ObjectNode. Must use an ObjectMapper to create them.
Can no longer use (object.get(key) instanceof String) – use (object.get(key).isTextual()) instead.

Relies on https://github.com/vivo-project/Vitro/pull/60


j2blake:
Checks failed with compile errors because vivo-project/Vitro#60 has not been applied.

PR #55, opened 2017-08-30
[VIVO-1374] Remove javascript disabling first name on forms

gneissone:
Addresses [VIVO-1374](https://jira.duraspace.org/browse/VIVO-1374)


grahamtriggs:
Thanks Benjamin - I've replaced the offending lines with the correct disabling code rather than removing, because whilst we haven't found a case where it's needed, it seems prudent to not run into one later.

PR #60, opened 2018-01-23
shorten ORCID reg-ex to increase legibility

katrinleinweber:
Hello!

Is there a specific reason your ORCID reg-ex are so long? The [`blame`](https://github.com/vivo-project/VIVO/blame/47f1408b3e9db4934c33f321689ef307b9cf4bdd/webapp/src/main/webapp/WEB-INF/ontologies/update15to16/sparqlConstructs/additions/orcid.sparql) didn't specify one, so I thought this suggested shortening might be useful. Probably, more refactoring would be possible. Let's discuss, if there was indeed no specific reason for the in-reg-ex redundancy.

Cheers :-)

### DIFF
--- a/webapp/src/main/webapp/WEB-INF/ontologies/update15to16/sparqlConstructs/additions/orcid.sparql
+++ b/webapp/src/main/webapp/WEB-INF/ontologies/update15to16/sparqlConstructs/additions/orcid.sparql
@@ -7,6 +7,6 @@ CONSTRUCT {
 } WHERE {
     ?s vivo:orcidId ?orcidString 
     FILTER(isLiteral(?orcidString))
-    FILTER (regex(str(?orcidString), "^[0-9][0-9][0-9][0-9]-[0-9][0-9][0-9][0-9]-[0-9][0-9][0-9][0-9]-[0-9][0-9][0-9]([0-9]|X)$"))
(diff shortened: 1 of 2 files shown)


j2blake:
Katrin - not my code, but usually when a awkward regex is used instead of an obvious one, it is because the obvious one was not valid for the particular regex processor. This code dates from VIVO release 1.6, and I would guess that the "regex" filter in the Jena ARQ library did not accept the simpler syntax (jena-arq 2.8.7).

Have you tested these changes to see that they work in the current VIVO? (jena-arq 3.4.0). Can you include a unit test to support the changes?

katrinleinweber:
I don't know Java. Or at least not enough to write a test :-/ Same with the build instructions https://wiki.duraspace.org/display/VIVODOC19x/Installing+VIVO#InstallingVIVO-GitHubInstallingfromGitHub. If the current tests (that Travis ran successfully) don't cover this change, please feel free to close this. No hard feelings and few sunken costs ;-)

j2blake:
Katrin; The current tests do not cover this change. I'll close the request.

You should know, however, that this pull request provided a major discussion point in yesterday's developer call. So, your pull request helps to move VIVO forward, even if not in the way that you expected ;-)

Thanks for helping to improve VIVO. -- Jim

PR #68, opened 2018-05-04
[VIVO-1447] Move ontology prefixes into separate file in filegraph directory

gneissone:
**[JIRA Issue](https://jira.duraspace.org/browse/VIVO-1447)**: VIVO-1447

Moves ontology prefixes from vitroAnnotations.n3 in the tbox/firsttime directory into a new file in tbox/filegraph.

# What's new?
Nothing new, just moved some triples from one file to another.

# How should this be tested?
* Reproduce the problem you are fixing (if applicable)
Upgrade VIVO with existing database ([instructions here](https://wiki.duraspace.org/display/VIVODOC110x/Upgrading+VIVO)) to develop branch. After upgrade, go to Site Admin > Ontology list. Prefixes are missing.
Missing prefixes screenshot

* Test that the pull request does what is intended.
Rebuild VIVO with this PR and follow steps above. All prefixes should be present.

# Additional Notes:
This only affects users upgrading.

# Interested parties
@mconlon17 @mjaved495

### DIFF
--- a/home/src/main/resources/rdf/tbox/filegraph/ontologies.owl
+++ b/home/src/main/resources/rdf/tbox/filegraph/ontologies.owl
@@ -5,8 +5,7 @@
      xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
      xmlns:owl="http://www.w3.org/2002/07/owl#"
      xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
-     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-     xmlns:vitro="http://vitro.mannlib.cornell.edu/ns/vitro/0.7#">
+     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
     <owl:Ontology rdf:about="http://xmlns.com/foaf/0.1/"/>
     <owl:Ontology rdf:about="http://purl.org/NET/c4dm/event.owl"/>
     <owl:Ontology rdf:about="http://vitro.mannlib.cornell.edu/ns/vitro/public"/>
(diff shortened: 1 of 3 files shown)


home/src/main/resources/rdf/tbox/filegraph/ontologies.owl
```diff
@@ -12,7 +12,7 @@
     <owl:Ontology rdf:about="http://purl.org/ontology/bibo/"/>
     <owl:Ontology rdf:about="http://vivoweb.org/ontology/core"/>
     <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/"/>
-    <owl:Ontology rdf:about="http://www.w3.org/2006/vcard/ns"/>
```
awoods: It appears that no substantive change is being made to this file. If that is true, please remove any changes to this file from the pull-request.

brianjlowe:
I don't agree with the assertion that nothing is new and that triples have simply moved from one file to another.   When triples are moved from firsttime/ to filegraph/, they cease to be editable by the end user via the GUI, so what is new is that the ontology prefix annotations are now uneditable.  Whether that's good or bad I don't know, but we should be explicit about it.

awoods:
Thanks @brianjlowe . 
@mjaved495 ? @gneissone ? Is it intentional and desired that the prefixes become uneditable by the user? Do users commonly (or ever) want/need to update the prefix mappings?

delsborg:
Why is this being done in the first place? I don't see the purpose stated in the pull request.

mjaved495:
I believe, @awoods  your comment was specifically on one file "ontologies.owl", that there is no change in this committed file. As far as moving triples from one file to another, yes it is a change. Preferably, no one should change the prefixes and I believe whoever (only admins can reach to that page, I believe) will be doing it, do understand the consequence of it.

mjaved495:
And I do not have knowledge, if application uses prefixes. My understanding is, where ever prefixes are used, URIs are also mentioned there.
Why it was required, because those who upgrade the VIVO installation from old version to latest, will having missing prefixes in "Ontologies List" page, unless they are in filegraph, instead of firsttime. firsttime folder is used only for fresh installations.

brianjlowe:
The ontology migration code should automatically merge in new additions to the firsttime annotations.  There shouldn't be a need to move things into filegraphs simply to handle upgrades.

awoods:
@brianjlowe : Can you point to "the ontology migration code"? It would be good to test that it works so that we can determine if this PR is obsolete.

brianjlowe:
And, unless the behavior has changed, if a user has edited an existing annotation value so that it differs from the values in the oldVersion directory, the new value will not be applied.  If the value is unchanged from its previous default, then a new default value will be automatically applied.  Of course, something entirely new that wasn't there before will just be added.

awoods:
As this conversation continues, I still believe that changes to home/src/main/resources/rdf/tbox/filegraph/ontologies.owl can be reverted from this PR.

gneissone:
@awoods, pulled back the non-changes. @brianjlowe I completely understand what you are saying. This change still allows admins to change ontology prefixes, but it overwrites the changes on every start which is definitely bad. 

I think the reasoning behind the PR is sound... ontology prefixes were previously included somewhere (filegraph?) and they were not included when the combined ontology and annotations files were created. /firsttime did not seem like a logical directory because it would result in the blank prefixes shown in the PR screenshot for any user upgrading. 

Keeping the prefixes in /firsttime and copying over missing ones upon upgrade using startup listeners and migration code seems ideal, but I don't have an understanding yet at this time.

mconlon17:
I'm probably missing something here, but an upgrade involves unloading the triples, creating a new database, and reloading the triples.  That *is* a firsttime condition and the prefixes from vitroAssertions.n3 will be loaded properly.  I'm not sure why this PR is needed.

gneissone:
Because we want to upgrade our old database and jena3tools requires both the content and config databases be empty when it runs. Once VIVO starts they are no longer empty. Plus we want to preserve our settings rather than reset everything to default by running firsttime procedures or putting editable settings in /filegraph (hence Brian's argument about moving the prefixes).

brianjlowe:
As I mentioned before, changes to firsttime are merged into an existing database as part of the ontology migration procedure.  You basically just need to create a 1.9-1.10 directory, copy and change the ASK query and success assertion, and include the oldAnnotations directory with a copy of firsttime from 1.9.  The upgrade will do the rest.

This should be a problem that was solved in 2010.  It doesn't seem like one worth re-creating unless we run out of things to do.

gneissone:
I agree with you. I'm closing this pull request since we are just talking in circles. Somebody needs to submit a new pull request with the 1.10.0 ontology migration code that will include the missing prefixes shown in the screenshot.

PR #73, opened 2018-06-25
handle doi url

Stefan-Wolff:
don't add "https://doi.org/" if the full URL is given

https://jira.duraspace.org/browse/VIVO-1527


gneissone:
screen shot 2018-06-25 at 9 44 11 am

Confirmed this is an issue. For ORCID, we support both full URLs and IDs alone, we should do the same for DOIs.

Stefan-Wolff:
of course

PR #74, opened 2018-06-25
handle doi url

Stefan-Wolff:
don't complete the URL if the full URL is given;

supported doi variants:
- http://doi.org/[doi]
- https://doi.org/[doi]
- http://dx.doi.org/[doi]
- https://dx.doi.org/[doi]
- doi:[doi]

https://jira.duraspace.org/browse/VIVO-1527

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-doi.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-doi.ftl
@@ -10,7 +10,12 @@
 <@showStatement statement property />
 
 <#macro showStatement statement property>
-    <#assign doi = (statement.value!).toLowerCase().replace("http://doi.org/", "").replace("https://doi.org/", "").replace("http://dx.doi.org/", "").replace("https://dx.doi.org/", "")>
+    <#assign doi = (statement.value!).toLowerCase().
+                    replace("http://doi.org/", "").
+                    replace("https://doi.org/", "").
+                    replace("http://dx.doi.org/", "").
+                    replace("https://dx.doi.org/", "").
+                    replace("doi:", "").trim()>
     <a href="https://doi.org/${doi}" title="${i18n().doi_link}" target="_blank">${doi}</a>
     <@lmt.addCitationMetaTag uri=(property.uri!) content=(doi) />
 </#macro>
--- a/webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-doi.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-doi.ftl
@@ -10,8 +10,9 @@
 <@showStatement statement property />
 
 <#macro showStatement statement property>
-    <a href="https://doi.org/${statement.value!}" title="${i18n().doi_link}" target="_blank">${statement.value!}</a>
-    <@lmt.addCitationMetaTag uri=(property.uri!) content=(statement.value!) />
+    <#assign doi = (statement.value!).replace("http://doi.org/", "").replace("https://doi.org/", "")>
+    <a href="https://doi.org/${doi}" title="${i18n().doi_link}" target="_blank">${doi}</a>
+    <@lmt.addCitationMetaTag uri=(property.uri!) content=(doi) />
 </#macro>
 
 
--- a/webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-doi.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-doi.ftl
@@ -10,7 +10,7 @@
 <@showStatement statement property />
 
 <#macro showStatement statement property>
-    <#assign doi = (statement.value!).replace("http://doi.org/", "").replace("https://doi.org/", "")>
+    <#assign doi = (statement.value!).toLowerCase().replace("http://doi.org/", "").replace("https://doi.org/", "").replace("http://dx.doi.org/", "").replace("https://dx.doi.org/", "")>
     <a href="https://doi.org/${doi}" title="${i18n().doi_link}" target="_blank">${doi}</a>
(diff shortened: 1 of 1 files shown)


gneissone:
For historical continuity, see https://github.com/vivo-project/VIVO/pull/74 for previous comments.

webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-doi.ftl
```diff
@@ -10,8 +10,14 @@
...
 <#macro showStatement statement property>
-    <a href="https://doi.org/${statement.value!}" title="${i18n().doi_link}" target="_blank">${statement.value!}</a>
-    <@lmt.addCitationMetaTag uri=(property.uri!) content=(statement.value!) />
+    <#assign doi = (statement.value!).toLowerCase().
+                    replace("http://doi.org/", "").
+                    replace("https://doi.org/", "").
+                    replace("http://dx.doi.org/", "").
+                    replace("https://dx.doi.org/", "").
+                    replace("doi:", "").trim()>
+    <a href="https://doi.org/${doi}" title="${i18n().doi_link}" target="_blank">${doi}</a>
```
gneissone: I think the modified DOI should only be used for the link. I would say revert to 'statement.value' for the text and the citation meta tag so the DOI is displayed in the format the user entered it in. 

[CrossRef guidelines](https://www.crossref.org/display-guidelines/#full) actually say to display the DOI as a link with https://doi.org/, but I say lets leave it up to the data owner to follow the rules.

Stefan-Wolff: Yes, i have thought on this. The problem is, that we are changing the link exactly to "https://doi.org/", but we are handling 5 different formats of the DOI URL. So we could get a link to "https://doi.org/" having a link text of "http://dx.doi.org/". This breaks with the user's expectation, when he clicks the link.
I think there are two possible solutions:
1. link href und link text are the same value
2. link text contains only the part of all handled DOI URLs -> the DOI
I prefer the second one.

gneissone: I see where you're coming from, but I think we should consider the data manager's expectations as well. If they enter the DOI in a certain format, I think they would expect it to be displayed the same way. The broken links are the big issue. I'm not sure if the average user is going to notice a discrepancy between what is displayed and the URL being followed if the end result is the same no matter the URL format.

I am in favor of displaying the DOI as entered (meanwhile fixing the link)... but I could also get behind reformatting the DOI to adhere to the Crossref guidelines, i.e. remove all the unapproved prefixes, then add https://doi.org to the display as you're doing with the link. But I don't agree with stripping all prefixes and displaying the DOI in a deprecated format.

Perhaps we can discuss during the developer call tomorrow.

Stefan-Wolff: If the data manager expect to see exactly the same link as he entered before, then he also expect to open exactly this link when he clicks it. 
Its a problem to show an URL and link to another URL. If only the DOI part is shown, then there would be no misinformation. 
#1 would also be a correct way to do it, but i think it looks not very nice to show such a long URL.

mconlon17: There are appear to be two issues:

1) fixing variants provided by users.  I think this is a good idea.
2) distinguishing between "DOI the URL" and "DOI the data value to be stored in the triple store".  I'm not sure there is such a difference.  The DOI as specified by CrossRef comes with https://doi.org/

brianjlowe: I'm in favor of displaying and linking according to the Crossref guidelines.  In any case, I would suggest that we not display one URL while linking to a different one in the href.  It's easy to image that some current or future browser (or plugin) might flag or disable such links as potential phishing attempts.

awoods: Supporting above comments, and noting that during [today's dev meeting](https://wiki.duraspace.org/display/VIVO/2018-06-26+-+VIVO+Development+IG), the recommendation is along the lines of option-1 in: https://github.com/vivo-project/VIVO/pull/74#discussion_r198024249

Namely, have the link and the display match, and for them both to be `https://doi.org/xxx`.

awoods: I believe the only change to this PR is in https://github.com/vivo-project/VIVO/pull/74/files#diff-19fb9fd9986b7f1f9d103022cdd1005dR19 , making the display the full `https://doi.org/xxx` URL.

tworrall: It is bad UX to change what a user enters in the background and without their knowledge. If VIVO expects the url to be in a specific format, and the user enters a different format, tell the user to fix it. That's the whole point of validation.

awoods: Agreed, @tworrall . The point that @gneissone made on today's call is that not all pathways to entering a DOI have validation (e.g. direct SPARQL-Update).

There is currently a bug. Would you be ok with the resolution proposed here to fix the bug, followed by another ticket that provides UI validation?

delsborg: at CU we don't use the editorial interface, we ingest via the harvester, and we only ingest the DOI number without the http[s]:// prefix, we count on VIVO for consistency. I would prefer the change made in VIVO.

tworrall: @awoods Yes.

PR #89, opened 2018-10-05
[VIVO-1605] Update to latest Error Prone plugin and configuration to support JDK 11

grahamtriggs:
**[VIVO-1605](https://jira.duraspace.org/browse/VIVO-1605)**: 

Updates the pom.xml to use the latest version of Google's Error Prone plugin and configuration, so that the build works with recently released JDKs

# What's new?
Nothing

# How should this be tested?
Do a full "mvn clean install" - the compilation should complete normally.

Should be tested with multiple JDK versions - should work for everything from JDK 8 through to the recently released JDK 11

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/installer/pom.xml
+++ b/installer/pom.xml
@@ -64,7 +64,7 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.5.1</version>
+                    <version>3.8.0</version>
                 </plugin>
                 <plugin>
                     <artifactId>maven-dependency-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


j2blake:
(duplicate of comment on https://github.com/vivo-project/Vitro/pull/88)

It seems inappropriate to use a SNAPSHOT version of a dependency. We could issue a VIVO release, and then a new SNAPSHOT of the dependency could be issued, potentially breaking our release.

Am I understanding this correctly?

roflinn:
Graham - Could you please update this to the release version of the plugin?  It was just released a few hours ago.  https://github.com/google/error-prone/releases/tag/v2.3.2

PR #90, opened 2018-10-08
has collaborator is symmetric

mconlon17:
# Add SymmetricProperty to has collaborator

[VIVO-1364](https://jira.duraspace.org/browse/VIVO-1364)

Adds a declaration that has collaborator is a symmetric object property.  This means if A has Collaborator B, then the inferencer can add B has collaborator A.

Adds four annotations to the data property to 1) define it; 2) give and example of usage; 3) provide the editor's name; 4) provide a citation for the definition.

# How should this be tested?
Before this pull request, declaring A has collaborator B does not establish B as a collaborator of A.
After the pull request, the inferencer adds B is a collaborator of A.

# Additional Notes
Sites should be alerted of this change.  While the change is in keeping with common use of the term "collaborator," the provision of a new assertion is a change for VIVO.

A socialization process defines collaboration.  A and B agree that they are collaborators.

# Interested parties
@VIVO-project/vivo-committers, @marijane, @vioil, @mjaved495, @hauschke, @brianjlowe

### DIFF
--- a/home/src/main/resources/rdf/tbox/filegraph/vivo.owl
+++ b/home/src/main/resources/rdf/tbox/filegraph/vivo.owl
@@ -1911,7 +1911,7 @@ there is a measurement process p that has specified output m, a measurement datu
     <!-- http://vivoweb.org/ontology/core#hasCollaborator -->
 
     <owl:ObjectProperty rdf:about="http://vivoweb.org/ontology/core#hasCollaborator">
-        <rdf:type rdf:resource="http://www.w3.org/2002/07/owl#ReflexiveProperty"/>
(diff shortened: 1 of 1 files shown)


mconlon17:
The inferencer *does not* add the reflexive assertion.  Does Vitro not support reflexive object properties?

mconlon17:
Thanks Brian.  Symmetric is the right idea, not reflexive.  I have updated and tested and everything works well including inference.

As for previous discussions, I believe this should be symmetric.  A and B should agree they are collaborators before either asserts they are.  I believe this is the normal usage of the term.  "Asymmetric collaboration" is not collaboration.  We might think of other terms for such relationships, and they often indicate mentoring, which is already covered.

Regardless, I believe the change here brings VIVO terminology in line with common use of the term "has collaborator".

PR #91, opened 2018-10-12
[VIVO-1607] Populate build number within installer directory

gneissone:
**[VIVO-1607](https://jira.duraspace.org/browse/VIVO-1607)**: ${buildNumber} does not populate

Resolves a problem where ${buildNumber} was never populated in revisionInfo.txt, therefore the app just displayed "${buildNumber}" instead of the actually hash code.

# What's new?
Runs the codemojo build number plugin and calculates the build date from within the installer directory.

# How should this be tested?
* Reproduce the problem you are fixing (if applicable)
Install VIVO without the changes. Login as a site admin. Click the the VIVO version number at the bottom of the page (or just go to http://localhost:8080/vivo/revisionInfo). The revision info will just say "${buildNumber}." 

* Test that the pull request does what is intended.
Install VIVO with changes. Do the same as above, the revision info should now have an actual hash value calculated from the git commit.

# Additional Notes:
The revision info worked with Vitro, but not with VIVO. I'm not exactly sure why this is. There might be a better, less naive way to accomplish this task if we can figure that out. Is it possible to install VIVO without the installer directory? If not, the plugin task and timestamp calculation can be removed from VIVO/pom.xml so they won't have to be performed twice. 

# Interested parties
@grahamtriggs @awoods

### DIFF
--- a/installer/pom.xml
+++ b/installer/pom.xml
@@ -16,6 +16,8 @@
         <vitro-version>${project.version}</vitro-version>
         <maven-site-plugin.skip>true</maven-site-plugin.skip>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.build.timestamp.format>yyyy-MM-dd HH:mm:ss</maven.build.timestamp.format>
+        <build.timestamp>${maven.build.timestamp}</build.timestamp>
     </properties>
 
     <build>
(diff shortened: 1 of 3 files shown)


mconlon17:
Seems like we should get a build number in all scenarios.  When I download source and build unchanged, I get no build number.  What should be done to correct this so I get a build number.

awoods:
Closed by: https://github.com/vivo-project/VIVO/commit/3af36c7cf45780f6d6978a82568d355d0fae9b74

PR #94, opened 2018-11-06
[VIVO-1619] Create weblinks for data properties that have a range of  anyURI and begin with http

delsborg:
[VIVO-1619](https://jira.duraspace.org/browse/VIVO-1619)

Identifiers that are data properties do not link to a URL even if the range is set to URL.
This change displays the data property value as a weblink IF it is of range URL AND it contains the string http.

# What's new?
Updated the propStatement-dataDefault.ftl file to test the range and value and create a weblink.
All data properties that have a range of URL AND have http in the value will display as a link.
There is no additional verification of the value. This validation should happen upstream in the VIVO ETL or as the user is entering the data.

Example:
* Changes https://github.com/delsborg to a hyperlink

# How should this be tested?
1. Deploy this change
2. Add some data to a data property with a range of anyURI
3. Verify that the data property is now a hyperlink.

# Additional Notes:
This has been tested in CU Boulders development instance.
I would like to see if this can also be applied to a maintenance patch in version 1.9.3 such that we can update our local VIVO repo from that and keep things in sync.

Example:
* Documentation regarding linking to external sties should be updated. It is currently sparse and difficult to identify how to link to external sites.
* Does this change add any new dependencies?  NO
* Does this change require any other modifications to be made to the repository? No
* Could this change impact execution of existing code?  NO


webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-dataDefault.ftl
```diff
@@ -15,7 +15,13 @@
 <@showStatement statement property datatype />
 
 <#macro showStatement statement property datatype>
+
     <#assign theValue = statement.value />
+
+<#-- create a weblink if the property range is a URI AND it contains http -->
+    <#if property.rangeDatatypeURI?? && theValue?contains("http")>
```
gneissone:
```suggestion
    <#if property.rangeDatatypeURI?? && (datatype == "anyURI") && theValue?contains("http")>
```
Line 22 is only checking if the property's range URI exists, not confirming that the range is xsd:anyURI. With the current logic, any default data property with a range, e.g. xsd:string that happens to also contain 'http' in its value will be turned into a link. Attached is an example of this problem with the property 'hasRestriction.'
screen shot 2018-11-06 at 2 58 52 pm

I think my suggested logic will work as intended based on my interpretation. Note that it seems like only vcard properties have the anyURI datatype out of the box. They already have special templates so unless a user changes the range for an existing property or adds a new property with anyURI as a range, there won't be any difference in function.

grahamtriggs: Use starts_with instead of contains so that it only matches where the protocol is http

webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-dataDefault.ftl
```diff
@@ -15,7 +15,13 @@
...
+        <#assign theValue = "<a href=\"" + statement.value + "\" target=\"_blank\">" + statement.value + "</a>" />
```
grahamtriggs: Would be good to add an "external" class to the link so that it can be decorated in CSS to indicate it is (probably) an external link.

grahamtriggs: Currently, no. But then there are also few places where external links are generated, and it's relatively easy for CSS designers to target those instances specifically to decorate them if desired.

By linking URIs based on the recognised datatype, they could potentially occur anywhere in the ontologies. As such, the only way someone could decorate them - and not the application URIs - is if there is a CSS class.

delsborg: @grahamtriggs I was looking into your point about making sure the changes are also in the Vitro file. I was wondering why there is a VIVO version. In examination it looks like there are only a few differences, all of which you made about 2 years ago in https://jira.duraspace.org/browse/VIVO-1237.  Can you explain why these changes can just be made in the Vitro file such that we can abandon the VIVO file?

grahamtriggs: The changes for the citation tags was not made in the Vitro version, because outputting the tags requires knowledge of the ontologies that you are working with (which Vitro can't assume).

However, on reflection it is probably the case that a lib-meta-tags.ftl with a no-op macro could be placed in Vitro, being overridden by the ontology aware version that is in Vivo, and the propStatement-dataDefault.ftl changes moved into Vitro.

delsborg: So it sounds like I should copy https://github.com/vivo-project/VIVO/blob/develop/webapp/src/main/webapp/templates/freemarker/lib/lib-meta-tags.ftl to the Vitro location and then do .... what -- change the switch statement? Can we just leave the cases in the switch as is? Not sure what the harm is for leaving it as is in Vitro, if the ontology isn't there the switches will never be fullfilled. Right?
Then, when we resolve above I can create a Vitro fork and test that in our 1.9.3 CU Boulder dev environment.
After that would I just copy the propStatement-dataDefault.ftl from Vivo to Vitro and then remove the file from VIVO?

webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-dataDefault.ftl
```diff
@@ -1,4 +1,4 @@
-<#-- $This file is distributed under the terms of the license in LICENSE$ -->
+<#-- $This file is distributed under the terms of the license in /doc/license.txt$ -->
```
gneissone:
```suggestion
<#-- $This file is distributed under the terms of the license in LICENSE$ -->
```
Undo reversion to old license location

webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-dataDefault.ftl
delsborg: can anybody confirm my steps above? THis would mean that lib-meta-tags.ftl in Vitro does contain elements from the VIVO ontologies, but I'm not sure this is a big deal except from a purist standpoint. It would eliminate code bloat and redundancy to do it this way. IMHO it's best shrink the code best instead of expanding it to reduce complexity.

webapp/src/main/webapp/templates/freemarker/body/partials/individual/propStatement-dataDefault.ftl
delsborg: Updated pull request to remove the propStatement-dataDefault.ftl. This is in Vitro with Vitro pull 101

delsborg:
Mike, it seems that fixing a range in the ontology is out of scope for this particular request, even though I think it makes sense at first glance. But I also think we must be cautious because we should be mindful of how current VIVO sites are using bibo:doi . This could be a "breaking change" to ingest and custom display. I recommend that your request be handled in a separate ticket and separate pull request. In addition the file change is being moved from the VIVO repo to Vitro which is unaware of the Vivo ontologies.

mconlon17:
Don,  I agree.  The two issues should not be combined.  #94 can not be tested using the existing VIVO ontology.  Testers should create a local ontology with a data property that has range anyURI.  I have updated the testing procedure on this pull request.

I have opened https://jira.duraspace.org/browse/VIVO-1662 to provide anyURI as the range for bibo:doi.

delsborg:
@mconlon17 see comment in Jira for additional comments regarding new data properties with anyURI range and why we did this at CU Boulder.
https://jira.duraspace.org/browse/VIVO-1619?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=61791

awoods:
@grahamtriggs : Do you still have concerns on this PR? Otherwise, I believe it is ready to merge.

PR #98, opened 2018-12-04
[VIVO-1340] Fix links to documentation in example.runtime.properties

lwrubel:
**[VIVO-1340](https://jira.duraspace.org/browse/VIVO-1340)**

Fixes links to documentation in example.runtime.properties

Removes broken links to version-specific documentation for properties. Adds link to top-level documentation page for all versions, with pointer to sections for configuration reference (recent versions) or installation properties (older versions).

# Additional Notes:
Assumes that the wiki will continue to have a page listing all documentation versions. 

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/home/src/main/resources/config/example.runtime.properties
+++ b/home/src/main/resources/config/example.runtime.properties
@@ -7,6 +7,11 @@
 # Save a copy of this file as runtime.properties in your VIVO home directory, 
 # and edit the properties as needed for your installation.
 #
+# For more information on specific properties, see the configuration reference
+# or installation options section of the technical documentation for the
+# version of VIVO you are running:
+# https://wiki.duraspace.org/display/VIVO/VIVO+Technical+Documentation 
+#
 # -----------------------------------------------------------------------------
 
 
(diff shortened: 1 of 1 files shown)


gneissone:
The Vitro version could use a similar treatment: https://github.com/vivo-project/Vitro/blob/develop/home/src/main/resources/config/example.runtime.properties

awoods:
@gneissone : In looking at: 
https://github.com/vivo-project/Vitro/blob/develop/home/src/main/resources/config/example.runtime.properties

It appears that the following link is "broken": https://wiki.duraspace.org/display/VIVO/Use+HTTP+caching+to+improve+performance
...but since it was not a tiny-url, it takes the client to a selection of potential pages, which seems useful. 

Do you have thoughts on the link above? ..and are there other links that should be updated in the Vitro `example.runtime.properties`?

/me wondering if we need to make a ticket, or if actually things are ok.

gneissone:
@awoods I suppose the same generic link to the technical documentation should be included in the Vitro runtime.properties. I thought the inline direct links to the pages describing the properties were useful, but would agree they aren't worth the effort to both keep them up to date and make sure changes to the wiki don't break the old versions.

PR #100, opened 2018-12-05
[VIVO-1528] annotate fields for HTML editing

Stefan-Wolff:
This pull request is part of https://github.com/vivo-project/Vitro/pull/76 .

### DIFF
--- a/home/src/main/resources/rdf/tbox/firsttime/vitroAnnotations.n3
+++ b/home/src/main/resources/rdf/tbox/firsttime/vitroAnnotations.n3
@@ -1730,7 +1730,9 @@ vivo:outreachOverview
       vitro:prohibitedFromUpdateBelowRoleLevelAnnot
               <http://vitro.mannlib.cornell.edu/ns/vitro/role#public> ;
       vitro:publicDescriptionAnnot
-              "Please enter a single summary narrative of your outreach goals and/or contributions"^^xsd:string .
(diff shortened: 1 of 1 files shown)


gneissone:
Can you please include the new statements in the statement blocks that already exist for each property?  There's no functional difference but all the other statements are already grouped for easy reference. For example, instead of having another vivo:overview block at the bottom, add it to the one that already exists:
```
vivo:overview
      vitro:editing
              "HTML"^^xsd:string ;
      vitro:displayLimitAnnot
              "1"^^xsd:int ;
      vitro:displayRankAnnot
              "10"^^xsd:int ;
      vitro:hiddenFromDisplayBelowRoleLevelAnnot
              <http://vitro.mannlib.cornell.edu/ns/vitro/role#public> ;
      vitro:hiddenFromPublishBelowRoleLevelAnnot
              <http://vitro.mannlib.cornell.edu/ns/vitro/role#public> ;
      vitro:inPropertyGroupAnnot
              <http://vivoweb.org/ontology#vitroPropertyGroupoverview> ;
      vitro:prohibitedFromUpdateBelowRoleLevelAnnot
              <http://vitro.mannlib.cornell.edu/ns/vitro/role#public> ;
      vitro:publicDescriptionAnnot
              "A short narrative summary to be used as a single descriptive overview statement."^^xsd:string .
```

mjaved495:
Have we defined the property vitro:editing in any file ?

mjaved495:
Noticed the commit in Vitro.... pls discard my above message.

PR #105, opened 2018-12-05
[VIVO-1656] Extension mechansim for passing properties to templates

grahamtriggs:
**[VIVO-1656](https://jira.duraspace.org/browse/VIVO-1656)**:

# What does this pull request do?
Uses the extension mechanism to pass the Plum Analytics and AltMetric options to the individual templates.

Requires the corresponding pull request from Vitro.

# What's new?
No change in functionality, this is an architectural improvement.

# How should this be tested?
Visit the page of any "individual" in VIVO. Should work exactly as it did before.

In particular, check publications that previously display Plum Analytics / AltMetric badges - these badges should still appear as before.

# Interested parties
@VIVO-project/vivo-committers

PR #107, opened 2018-12-05
[VIVO-1415] Add publication claiming from PubMed and CrossRef

grahamtriggs:
**[VIVO-1415](https://jira.duraspace.org/browse/VIVO-1415)**: 

Implements the DOI and PMID claiming process from OpenVIVO

# What's new?
Adds the claiming interface - links to claim by DOI or PMID and the additional pages for entering IDs and confirming the links.

Most of the code is self-contained in the package org.vivoweb.webapp.createandlink;

Only small changes to the OpenVIVO code has been made, in order to move CSS classes to the appropriate theme files, add the claiming links into both the Wilma and Tenderfoot themes, and to make URIs conform to the standard VIVO format (as using the URIs generated in OpenVIVO would require further customisation of VIVO).

Also, there are contribution options, as the contribution ontology is not formally adopted into VIVO / we haven't considered how this applies to publications that aren't sourced via the claim interface, etc.

# How should this be tested?
Log in to VIVO with the ability to edit.
Go to a profile and verify the options for claiming are there.
Click on the claiming links - you should be claiming articles for the profile that you came from
Enter IDs, check that the publications are returned as expected, and confirming creates new publications and user links (note that if the DOI or PMID is already present, then the claim interface will return the existing publication from VIVO, not the external metadata).
Go to a co-author profile and claim the publication for them, verifying that the publication now links them as an author.

# Additional Notes:
This patch relies on the changes in this ticket / pull request:
https://jira.duraspace.org/browse/VIVO-1656
Vitro PR: https://github.com/vivo-project/Vitro/pull/99
VIVO PR: https://github.com/vivo-project/VIVO/pull/105

# Interested parties
@VIVO-project/vivo-committers


roflinn:
I have recently been looking at the Scopus API it may be a good addition to this set in the future.

webapp/src/main/webapp/i18n/vivo_all.properties
```diff
@@ -810,3 +810,53 @@ role_in_presentation_capitalized=Role in Presentation
...
+create_and_link_unlisted_author=Unlisted Author
+create_and_link_editor=Editor
+create_and_link_not_mine=This is not my work
+create_and_link_remaining=There are {0} ids remaining
+create_and_link_thank_you=Thank you
+create_and_link_finished=There are no more works left to claim.<br />You may enter more IDs below, or view your profile.
+create_and_link_go_profile=Go to profile
+create_and_link_enter_dois_intro=You may enter one or more DOIs to match, and can be entered either as an ID or URL:<br /><br />e.g.
+create_and_link_enter_dois_supported=Currently, DOIs issued by Crossref, DataCite and mEDRA are supported.<br />Each DOI should be separated by a comma, semi colon or new line.
+create_and_link_enter_pmid_intro=You may enter one or more PubMed IDs to match. Each ID should be separated by a comma, semi colon or new l
```
gneissone:
```suggestion
create_and_link_enter_pmid_intro=You may enter one or more PubMed IDs to match. Each ID should be separated by a comma, semi colon or new line.
```

gneissone:
> I have recently been looking at the Scopus API it may be a good addition to this set in the future.

or the Web of Science API 😜

roflinn:
"What if somebody else has already claimed your authorship? Currently it's visible, but not selectable with no recourse. We hopefully won't have to support this situation. I would want to place a link to 'contact us' on the claiming pages for questions if this were to ever arise, however."

Would this be an issue for multiple authors?

gneissone:
> "What if somebody else has already claimed your authorship? Currently it's visible, but not selectable with no recourse. We hopefully won't have to support this situation. I would want to place a link to 'contact us' on the claiming pages for questions if this were to ever arise, however."
> 
> Would this be an issue for multiple authors?

Multiple authors on a paper are allowed, but only one profile may be linked to an individual author name. Which makes sense, that's how it should be. Just imagining the worse case scenarios that could happen if we enforce that rule.

grahamtriggs:
@gneissone The changes in PR 105 (Vitro#99) are required for this PR. In order to cherry-pick the changes from our GitLab, I created the branch for this PR off of the branch for #105. So you are seeing the same *commit* in the history that provides those changes - this will resolve itself when #105 is merged.

roflinn:
basically you pulled the code out of vitro and moved to over to vivo since it makes more sense there if i'm reviewing the commits correctly

grahamtriggs:
@roflinn That's correct - it didn't make any sense that AltMetrics and Plum Analytics was enabled in Vitro, when it is relevant only to the VIVO domain.

However, it's a bigger issue with this code - because I want to pass configuration to say whether DOI or PMID claiming is enabled, As I wanted to call code that is in the VIVO layer to do that, it couldn't be done in Vitro. And yes, you can completely turn off DOI/PMID claiming from the configuration, in the case that you have other authoritative publication sources and don't want users adding their own.

grahamtriggs:
@gneissone Thanks for spotting the PMC ID issue - that's now resolved.

I've also normalised the DOIs to insert https://doi.org/....

Editorship is now checked and won't let you claim an article you are already an editor for.

An error message is displayed if no citation could be retrieved.

BTW - you might have missed the configuration element in the runtime.properties - you can configure whether claiming is available or not from there.

gneissone:
Thanks for the fast fixes, I will give them a spin. You're right, I missed the setting in runtime.properties. Your explanation of how names are displayed after it has been claimed, and how an author with a 'stolen' citation can still add the pub, make sense.

Minor gripe, /doc/license.txt doesn't exist anymore, the licensing text is now
`/* $This file is distributed under the terms of the license in LICENSE$ */`

grahamtriggs:
@gneissone Thanks for spotting that. I had noticed similar, but thought it was likely down to other theming CSS changes I had made. I've updated the CSS to prevent this happening.

Also, I've taken the opportunity to update the way the DOI example is presented.

BTW, DataCite has been having issues with it's services over the past few days. So at the moment claiming DataCite DOIs doesn't work, because the metadata isn't being returned via the negotiator as it should be.

awoods:
@grahamtriggs : If you can rebase this on `develop`, we can move it forward. Thanks!

awoods:
@grahamtriggs : Is it correct to assume that the first commit in this pull-request does not belong?
https://github.com/vivo-project/VIVO/pull/107/commits/f2021caebb07d6c78de4cc8281f703359582c14a

grahamtriggs:
@awoods No, it's a bit more complicated. I've rebased, and GitHub is still showing this PR to have more commits than are in branch, so I'll close this and re-open from the branch.

PR #108, opened 2018-12-05
[VIVO-1287] Reclassify five classes that are subClass skos:Concept

mconlon17:
**[VIVO-1287](https://jira.duraspace.org/browse/VIVO-1287)**

Provides new parent classes for five classes.  These five classes were classed as skos:Concept causing VIVO to return instances in the list of Concepts, an undesirable behavior, because the classes are not skos:Concepts

# What's new?
* AcademicDegree classified as EducationalProcess
* Award classified as Process
* Credential classified as Process
* ARG_2000376 classified as Quality
* DateTimePrecision classified as Quality

# How should this be tested?
* Build (no need to clean out data)
* Review existing instances of degrees, awards, credentials, and dates -- each should display and edit normally.  ARG_2000376 is not used by the VIVO application, no testing is needed.
* Create new degrees, awards, credentials and dates.  Each should create and display normally.
* View the Index to see that instances of these classes no longer appear in the index.

# Additional Notes:
1. The Class diagram in the VIVO Technical documentation will need to be updated when software is released.  The ontology diagrams should not need to be updated -- there is no change in the models, only the superclass is effected.
1. Sites with existing instances of these classes (that's all existing VIVO sites) may wish to implement CONSTRUCTs to update the parent classes for existing instances (these can be provided so that sites do not have to each write them) when this software is released.  New instances will not appear in the concept list.  Old instances will continue to appear until CONSTRUCT is run.

# Interested parties
@VIVO-project/vivo-committers @vivo-project/contributors @vivo-project/vivo-ontologists

### DIFF
--- a/home/src/main/resources/rdf/tbox/filegraph/vivo.owl
+++ b/home/src/main/resources/rdf/tbox/filegraph/vivo.owl
@@ -5954,7 +5954,13 @@ To enable other Gender/Sex codes to be used, this dataproperty has range URI. Th
 
     <owl:Class rdf:about="http://purl.obolibrary.org/obo/ARG_2000376">
         <rdfs:label xml:lang="en">Contact Qualifier</rdfs:label>
-        <rdfs:subClassOf rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
+        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/BFO_0000019"/>
+        <obo:IAO_0000112 xml:lang="en">Examples are business address, shipping address.</obo:IAO_0000112>
+        <obo:IAO_0000114 rdf:resource="http://purl.obolibrary.org/obo/IAO_0000120"/>
+        <obo:IAO_0000115 xml:lang="en">The purpose of contact information.</obo:IAO_0000115>
+        <obo:IAO_0000116 xml:lang="en">Contact qualifiers are represented as individuals of this class.  We do
+        not have an authoritative controlled vocabulary for contact qualifiers.</obo:IAO_0000116>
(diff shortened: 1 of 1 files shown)


brianjlowe:
While the change to ARG_2000376 and DateTimePrecision seem less controversial, I don't like the idea of making the other three subclasses of Process.  While we might prefer that a future version of the ontology might model things using these kinds of processes, such a change to the current TBox would put it out of sync with the real-world ABox data in present-day VIVOs.

After reflecting a bit I remember the 2013 discussions that let to these classes being placed under skos:Concept, and I still agree with the reasoning.  The fundamental problem is that the instances, for example, of vivo:Award should really be classes and not individuals.  The URI for the vivo:Award instance labeled "John A. and Mary Smith Award for Excellence" is really the identifier of a group of individual awards issued to particular persons on particular dates (called AwardReceipts in VIVO) that share certain properties in common.  That URI should really be the URI of a class.  Converting such individuals into TBox entities, however, was deemed a step too far even given the relatively radical changes put forth in the ISF, so the next best thing was to use SKOS, which is basically a way of turning what would be classes in a "real" ontology into instances of a single class.  Thus, Award, AcademicDegree and Credential became subclasses of skos:Concept.

If these axioms cause problems in application use, might I propose as a compromise that they simply be omitted from the VIVO application and Award, AcademicDegree and Credential allowed to float up to the top under owl:Thing where we might hope they cause less harm?  Sites wishing to retain the old hierarchy could add the axioms back to their own local TBox, and wouldn't have to remove a new addition.

delsborg:
Brian, I agree with your points regarding subclassing these under process for current VIVO's. It might be accurate, and I see that WIKIDATA does this, but I'm not sure what this will do to our current VIVO instance.
 At CU we just have the parent class omitted thus they are just owl.Things. So for the sake of a small ontological fix your compromise sounds appropriate.

PR #111, opened 2018-12-07
[VIVO-1662] Add range anyURI to bibo:doi

mconlon17:
[JIRA Issue](https://jira.duraspace.org/browse/VIVO-1662)

Modifies the VIVO ontology to add the range anyDOI to bibo:doi

# What's new?
Crossref has recommended that doi be displayed as doi since 2011.  In 2017 they simplified their recommendation.  You can find it here:  https://www.crossref.org/display-guidelines/

In VIVO currently there is no range declaration on bibo:doi.

This change should have no impact on existing VIVO systems, nor on their data.  VIVO does not currently enforce ranges

Coupled with [VIVO-1619](https://jira.duraspace.org/browse/VIVO-1619), and doi recorded using the CrossRef recommendation, VIVO will provide links automatically to the landing pages of all entities with doi meeting the CrossRef recommendation.

# How should this be tested?
1. Build, no need to empty the database.  Ontology is in a filegraph and will be updated automatically.
1. View the bibo:doi datatype property via Site Admin > Datatype Property Heirarchy > All Datatype Properties > DOI.
1. You should see range datatype http://www.w3.org/2001/XMLSchema#anyURI

# Additional Notes:
1. If/when [VIVO-1619](https://jira.duraspace.org/browse/VIVO-1619) is merged, DOI stored according to the CrossRef recommendation will appear as clickable link
2. Sites may optionally wish to update their DOIs in their data to increase the number of DOI that meet the CrossRef recommendation and are available as clickable links (assuming 1619).  This can be done using CONSTRUCT statements in SPARQL along with Add and Remove RDF.  The VIVO project may wish to provide sample queries for this purpose.

Example:
* Does this change require documentation to be updated? YES, when a release including this change is public
* Does this change add any new dependencies? NO
* Does this change require any other modifications to be made to the repository? NO.  The change described above is optional.  Sites that do not modify their data will have less clickable links.
* Could this change impact execution of existing code? NO

# Interested parties
@VIVO-project/vivo-contributors @vivo-project/vivo-ontologists

### DIFF
--- a/home/src/main/resources/rdf/tbox/filegraph/vivo.owl
+++ b/home/src/main/resources/rdf/tbox/filegraph/vivo.owl
@@ -4104,7 +4104,16 @@ Definition and description came from Wikipedia here: http://en.wikipedia.org/wik
         <ns:term_status>stable</ns:term_status>
         <rdfs:comment xml:lang="en">Digital Object Identifier</rdfs:comment>
         <rdfs:domain rdf:resource="http://purl.obolibrary.org/obo/IAO_0000030"/>
+        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#anyURI"/>
         <rdfs:subPropertyOf rdf:resource="http://purl.org/ontology/bibo/identifier"/>
+		<obo:IAO_0000112 xml:lang="en">OpenVIVO: Transparency in Scholarship has DOI https://doi.org/10.3389/frma.2017.00012</obo:IAO_0000112><!-- example of usage -->
+		<obo:IAO_0000114 rdf:resource="http://purl.obolibrary.org/obo/IAO_0000120"/><!-- curation status -->
+		<obo:IAO_0000115 xml:lang="en">In computing, a Digital Object Identifier or DOI is a persistent identifier or handle used to uniquely identify objects, standardized by the International Organization for Standardization (ISO). https://en.wikipedia.org/wiki/Digital_object_identifier</obo:IAO_0000115><!-- definition -->
+		<obo:IAO_0000116 xml:lang="en">VIVO is adding a range declaration of anyURI to bibo:doi to move toward the CrossRef recommendation (2017) for displaying DOI as URI. As URI, Vitro and VIVO will render them as clickable links, providing access to the object's landing page for all objects with doi.</obo:IAO_0000116><!-- editor's administrative note -->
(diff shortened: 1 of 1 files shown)


mjaved495:
I am not in favor of changing definition of an external entity. This data property is defined in BIBO ontology and BIBO ontology has a specified range for this property (line #1024 in https://github.com/structureddynamics/Bibliographic-Ontology-BIBO/blob/master/bibo.owl)

I would be "ok" if the defined range (xsd:anyURI) is not part of vivo.owl ontology but within an application graph that is specific to VIVO application only.

mconlon17:
Agree with everything said here.  Was unaware there was a range on bibo:doi -- the VIVO ontology copy did not have that, there was no range assertion.  We could, of course, easily create vivo:doi

Yes for working with the developers of bibo, but as we all know, this sometimes works and sometimes (often) does not.  This is a downside to using other people's ontologies.  @vioil has recently asked them about the license for their ontology.

The issues do not arise with the Identifer proposal.  The identifier could have a value, a URL and other properties as needed. And would be ours to modify as needed.

Inconsistency is unfortunately common.  Changing to URL is possible, and could be "required" -- provided as an upgrade script for consistency.

gneissone:
Since commit https://github.com/vivo-project/VIVO/commit/1343966884bedb738772f0f537cbbf5c44ab8c15, we parse DOI values to display them in the recommended format automagically. There was a fair amount of discussion about whether this was appropriate or not. I think my opinion was slightly off from the eventual behavior, but I prefer the current solution to creating a new vivo:doi property or changing the range when bibo:doi already has one. 

We already have logic in place to support DOIs in both the old and new (and a few other) formats. I think a better solution to accomplish what I think is this PRs intention is to add validation to the form itself to force DOIs entered through the UI to be the preferred format, but still store them as a string to conform with the bibo:doi range.

We should also petition whoever is in charge of the bibo ontology to change the range :)

brianjlowe:
The rdfs:range axiom in BIBO is basically redundant:  that the values are rdfs:Literals is implied already by the fact that the property is declared as a DatatypeProperty.  There is nothing wrong with typing bibo:doi values in individual statements with the datatype xsd:anyURI, as all xsd:anyURI values are also rdfs:Literal values.

mconlon17:
After some very thoughtful comments from many people, this pull request will not move forward.

Perhaps there are three reasons and I many not state them as clearly as other people might:
1. Modifying the definition of a term we are using from another ontology is a form of "ontology hijacking."  We don't hijack the ontologies of others.
1. The domain model was correct as is.  There was no need to change it. A DOI *does not* have range anyURI.  There are many ways DOI can be recorded.
1. Hints to the software belong elsewhere.
1. Data managers may wish to standardize on one form of DOI and that form may vary from site to site.

VIVO strives for full separation of the domain model from the software.

PR #125, opened 2019-04-22
[VIVO-1687] Set Tenderfoot pub chart height

gneissone:
**[JIRA Issue](https://jira.duraspace.org/browse/VIVO-1687)**: https://jira.duraspace.org/browse/VIVO-1687

https://stackoverflow.com/questions/12830897/svg-renders-but-gets-cut-off-in-firefox-only-why

Resolves an issue where the Tenderfoot publication bar chart is cropped in Firefox 

# What's new?
Sets the SVG width in javascript explicitly for Firefox compatibility

# How should this be tested?
Change theme to Tenderfoot. Look at bar chart in Firefox. It should be cropped in the vertical direction.
Screen Shot 2019-04-18 at 2 43 24 PM

Redeploy with changes. It should show the whole chart.

# Interested parties
@grahamtriggs

### DIFF
--- a/webapp/src/main/webapp/themes/tenderfoot/templates/body/partials/individual/individual-visualizationFoafPerson.ftl
+++ b/webapp/src/main/webapp/themes/tenderfoot/templates/body/partials/individual/individual-visualizationFoafPerson.ftl
@@ -9,14 +9,14 @@
 <#assign obo_RO53 = "http://purl.obolibrary.org/obo/RO_0000053">
 
 <#assign isInvestigator = ( p.hasVisualizationStatements(propertyGroups, "${obo_RO53}", "${core}InvestigatorRole") ||
-                            p.hasVisualizationStatements(propertyGroups, "${obo_RO53}", "${core}PrincipalInvestigatorRole") || 
+                            p.hasVisualizationStatements(propertyGroups, "${obo_RO53}", "${core}PrincipalInvestigatorRole") ||
                             p.hasVisualizationStatements(propertyGroups, "${obo_RO53}", "${core}CoPrincipalInvestigatorRole") ) >
 
 <#if (isAuthor || isInvestigator)>
- 
+
     ${stylesheets.add('<link rel="stylesheet" href="${urls.base}/css/visualization/visualization.css" />')}
     <#assign standardVisualizationURLRoot ="/visualization">
-        
+
         <#if isAuthor>
             ${scripts.add('<script type="text/javascript" src="${urls.base}/js/d3.min.js"></script>')}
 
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/themes/tenderfoot/templates/body/partials/individual/individual-visualizationFoafPerson.ftl
```diff
@@ -47,6 +47,8 @@
                             height = (chartHeight - margin.top - margin.bottom),
                             g = svg.append("g").attr("transform", "translate(" + margin.left + "," + margin.top + ")");
 
+                    svg.attr("width", chartWidth).attr("height",chartHeight)
```
awoods: I assume that setting the `width` and `height` in lines 46 and 47 does not work?

webapp/src/main/webapp/themes/tenderfoot/templates/body/partials/individual/individual-visualizationFoafPerson.ftl
```diff
@@ -9,22 +9,22 @@
 <#assign obo_RO53 = "http://purl.obolibrary.org/obo/RO_0000053">
 
 <#assign isInvestigator = ( p.hasVisualizationStatements(propertyGroups, "${obo_RO53}", "${core}InvestigatorRole") ||
-                            p.hasVisualizationStatements(propertyGroups, "${obo_RO53}", "${core}PrincipalInvestigatorRole") || 
+                            p.hasVisualizationStatements(propertyGroups, "${obo_RO53}", "${core}PrincipalInvestigatorRole") ||
```
awoods: :(

webapp/src/main/webapp/themes/tenderfoot/templates/body/partials/individual/individual-visualizationFoafPerson.ftl
gneissone: My guess is that those lines set the height and width of the visualization itself, but not the container it sits in.

dofeldsc:
I checked it and it is working for the height but if i use firefox then i have a certain width where 2018 and 2019 disappear.
vivo_publicationchart_widthproblem

But i have some local changes and had not the time to test it with a clean tenderfoot at this time. If nobody has this problem you can probably ignore my comment.

gneissone:
Can you please test it without your changes? 
I am not experiencing the cropping you describe, even with different resolution screens. But I am only testing on the latest version of Firefox in OS X so that's not a very big sample size.

Screen Shot 2019-04-26 at 9 06 10 AM

dofeldsc:
As i thought, it was a problem with my local changes :-/ Without my changes it looks like this in this special case so everything is fine. I found my useless css and removed it. Apologies for the delay of the ticket 
vivo10

PR #130, opened 2019-06-26
[VIVO-1670] ORCID - Disable Step 2 and Internationalization

grahamtriggs:
**[VIVO-1670](https://jira.duraspace.org/browse/VIVO-1670)**: 

Internationalization and ability to disable step 2 of ORCID interface in configuration.

# What's new?
Adds an option to example.runtime.properties for setting whether it is the "public" or "member" API that you have access to.

Setting this to "public" disables the Step 2 of integration, which uses the member's API to add a link to a person's ORCID profile.

Also, strings have been pulled out of the interface for internationalization.

# How should this be tested?
Set up a VIVO with ORCID integration.

Use without the new setting in runtime.properties, or with the setting orcid.apiLevel = member, the ORCID integration should appear and behave exactly as before.

Setting runtime.properties to orcid.apiLevel = public (and restarting) will result in only Step 1 being visible, and it does not attempt to add a link to your ORCID profile.

# Interested parties
@VIVO-project/vivo-committers


webapp/src/main/webapp/i18n/vivo_all.properties
```diff
@@ -810,3 +810,46 @@ role_in_presentation_capitalized=Role in Presentation
 advisee_capitalized_first_name=First Name
 advisee_capitalized_lastname=Last Name
 
+add_orcid_id=Add an iD
```
awoods: These should also be included in https://github.com/vivo-project/VIVO-languages, no?

gneissone: Created a pull request using the English i18n strings: https://github.com/vivo-project/VIVO-languages/pull/21

HACKHNATON01:
please Follow My Intruction...please write
https://www.google.com/url?sa=t&source=web&rct=j&url=https://www.vivo.com/en/support&ved=2ahUKEwivxvbmibjkAhVXiHAKHW-rDNQQFjAAegQIAxAB&usg=AOvVaw1yFn-PMJZmTn4O-X3Wa-9w

PR #136, opened 2019-07-26
VIVO-1614 - Update to allow the TPF API to be toggled via the runtime.properties.

roflinn:
Update to allow the TPF API to be toggled via the runtime.properties.


awoods:
Related to: https://jira.duraspace.org/browse/VIVO-1614

PR #142, opened 2020-01-20
[VIVO-1737] Update Error Prone

grahamtriggs:
**[VIVO-1737](https://jira.lyrasis.org/browse/VIVO-1737)**:

Updates the Error Prone plugin to for compile time analysis to the latest release, and moves the configuration to a profile that is enabled for JDK 1.8 - 11.

Builds with Error Prone fail when using JDK 12+, so by not enabling on those JDKs enables people to use them for their environment - Vitro/VIVO builds on JDK 13, but it is not tested.

This also allows developers to disable the Error Prone plugin from the command line, however I would not recommend this in general as it quickly and easily catches bad code.

# What's new?
Nothing specific, although the Error Prone plugin is now faster and includes some extra checks

# How should this be tested?
Run mvn clean install - build should complete.
Ideally, test with multiple JDK versions

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,6 +229,34 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <!-- Profile for Error Prone plugin -->
+            <id>errorprone</id>
+            <activation>
+                <jdk>[1.8,12)</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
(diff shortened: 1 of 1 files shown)


awoods:
Error Prone runs and VIVO builds as expected on Java 11.

PR #152, opened 2020-03-27
Sprint i18n whitespace

nicalico:
Fixed whitespace/tab indentations.


awoods:
It looks like you have gone above and beyond, @nicalico . Thank you.
You will want to target this pull-request to `sprint-i18n-whitespace` instead of the current target of `master`.

PR #170, opened 2020-05-20
Made modifications to publicationToPersonUtils.js so that multilangua…

matthiasluehr:
…ge support will not break the javascript functionalty.

**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1739)**

This pull request is replacing publicationToPersonUtils.js. In the original JS the fields are no longer
displayed dynamical if the language is different from English. 

# What's new?
In the old version the text value of the drop down list is used to deteremine what class is selected.
Translating the text values of the drop down menu breaks this code. In the new version the value
of the input (the URI of the class) is used to do this. 

# How should this be tested?
Go to a person's profile page, chose the selected publications tab and add a publication. Change the
type between 'Academic Article' and 'Book' and check if the input fields are changing. Switch to
another language than English and repeat this step.

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/edit/forms/js/publicationToPersonUtils.js
+++ b/webapp/src/main/webapp/templates/freemarker/edit/forms/js/publicationToPersonUtils.js
@@ -146,9 +146,10 @@ var publicationToPersonUtils = {
         // hide everything, then show what's needed based on type
         // simpler in the event the user changes the type
         this.hideAllFields();
-        var selectedType = this.typeSelector.find(':selected').text();
+        var selectedType = this.typeSelector.find(':selected').val();
+		alert(selectedType);
 
-        if ( selectedType == 'Academic Article' ) {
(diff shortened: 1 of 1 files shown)


awoods:
This looks like a very reasonable resolution... and works as expected for English and German. However, it appears that `fr-CA` is not working. It appears to be due to the fact that there is no `fr_CA` "initialTBoxAnnotations...nt" file:
```
./en_US/home/src/main/resources/rdf/tbox/everytime/initialTBoxAnnotations_en_US.nt
./pt_BR/home/src/main/resources/rdf/tbox/everytime/initialTBoxAnnotations_pt_BR.nt
./es/home/src/main/resources/rdf/tbox/everytime/initialTBoxAnnotations_es.nt
./de_DE/home/src/main/resources/rdf/tbox/everytime/initialTBoxAnnotations_de_DE.nt
```

Would you mind either adding the missing `fr_CA` file here, or creating a JIRA to do as a follow-on task?

PR #173, opened 2020-06-24
i18n: title tag and headings in capability map are in english in french version

dofeldsc:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1847)**

**[Second required PR](https://github.com/vivo-project/VIVO-languages/pull/56)**

The PR will add multi-language support for a part of the capability map and the title of the site.

# What's new?
This PR will add two variables for the multi-language support in the needed js-file and use them. If they are not provided by the languashould this be tested?
You go to the capability map, search for something (with sample data for example "Rhetoric") and then click the "informations" Button. The Group and the Term should be multi-lingual. Just like the title of the page.

# Additional Notes:
* the multi-language tags come with another PR in the VIVO-languages Repository

**Note:** Ignore the name of the branch, i accidentally used the number of my other ticket. 1847 would be the right onege-templates the previous terms ("Term" and "Group") will be used.

# How should this be tested?
You go to the capability map, search for something (with sample data for example "Rhetoric") and then click the "informations" Button. The Group and the Term should be multi-lingual. Just like the title of the page.

# Additional Notes:
* the multi-language tags come with another PR in the VIVO-languages Repository

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/visualization/capabilitymap/capabilityMap.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/visualization/capabilitymap/capabilityMap.ftl
@@ -15,6 +15,10 @@ ${stylesheets.add(
 )}
 
 <script language="JavaScript" type="text/javascript">
+    var i18nStringsCap = {
+        term: '${i18n().group_capitalized}',
+        group: '${i18n().term_capitalized}'
+    };
     var contextPath = "${urls.base}";
     $(document).ready(function() {
         var loadedConcepts = $.ajax({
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/templates/freemarker/visualization/capabilitymap/capabilityMap.ftl
```diff
@@ -15,6 +15,10 @@ ${stylesheets.add(
 )}
 
 <script language="JavaScript" type="text/javascript">
+    var i18nStringsCap = {
+        term: '${i18n().group_capitalized}',
```
awoods, line 19: These appear to be swapped: `term` is being populated by "group_capitalized".

dofeldsc: You are right, i changed it

dofeldsc:
> It is likely a configuration issue on my side, but I am unable to get the capability map to work in any non-English language :(

Do you see an error in the console of the borwser?

PR #186, opened 2020-08-28
Remove files replicated in i18n

awoods:
Related to: https://jira.lyrasis.org/browse/VIVO-1917

See: https://github.com/vivo-project/VIVO-languages/pull/67

No functional changes in the update... just reorganizing files.
Will be necessary to completely remove your triplestore and home/rdf files to test this update.


gneissone:
Perhaps I spoke too soon. I'm running into an issue with a completely fresh install now... the menu only shows 'Home' and the english i18n labels are missing upon initial startup... but if I restart the application they appear.

Initial state
Screen Shot 2020-09-02 at 11 54 53 AM

After restart 
Screen Shot 2020-09-02 at 11 55 25 AM

awoods:
Thanks for the review and catch, @gneissone . I have re-added `menu.n3`.

gneissone:
@awoods should the label and linkText literals be removed from menu.n3 since they are replicated in VIVO-languages?

awoods:
> @awoods should the label and linkText literals be removed from menu.n3 since they are replicated in VIVO-languages?

@gneissone : I was wondering about that as well. Since they do not have any `@lang` tags, it seems that they will always be superseded by another value. As you can see, I was inclined to leave them there... but could be persuaded otherwise.

gneissone:
If they are always superseded by the strings in VIVO-languages, and VIVO-languages is always required, I would argue for them to be removed. Why keep them if they're never used?

awoods:
PR updated based on code review comments.

gneissone (approved):

PR #192, opened 2020-09-29
[VIVO-1867] i18n: person's domain appears in English in French capability map

dofeldsc:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1867**

This Pr adds i18n-support for a person's domain in the capability map. For problem/bug description see Jira ticket.
The new SPARQL-Query will look for a appropriate label related to the selected language in this order:
1. label with the selected language + country (like "en-US")
2. label with the selected language (like "en")
3. label with the selected language and no matter which country (like "en-*", for example "en-CA")
4. any label no matter which langauge

# How should this be tested?
With the sample data (https://raw.githubusercontent.com/vivo-project/sample-data/master/sample-data.n3) go to "/vivo/organizations" and pick something (for example englisch) and add labels for some languages.
Then move to "/vivo/vis/capabilitymap" and search for something (for example "Rhetoric", the label needs to be set in every langauge).
If you click on the two persons in the cap-map, you see the information on the right site. This information under the name of the person (for example "Patricia Roberts, Professor and Chair", "englisch") should now support multiple langauges, for which you have previously set the labels.
For better explanation see the images in the Jira ticket.

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/visualization/utilities/VisualizationCaches.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/visualization/utilities/VisualizationCaches.java
@@ -234,13 +234,37 @@ final public class VisualizationCaches {
                     new CachingRDFServiceExecutor.RDFServiceCallable<Map<String, String>>(visualizationAffinity) {
                         @Override
                         protected Map<String, String> callWithService(RDFService rdfService) throws Exception {
+                            // get current selected language tag
+                            VitroRequest vreq = rdfService.getVitroRequest();
+                            String langCtx = "en-US"; // set default
+                            String language = "en"; // set fallback language
+                            try {
+                                langCtx = language = vreq.getLocale().getLanguage();
+                                if (!vreq.getLocale().getCountry().isEmpty()) {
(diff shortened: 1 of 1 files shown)


awoods (approved):
Works great.

It is possible that "[conceptToLabel](https://github.com/vivo-project/VIVO/blob/78f042016ea198dae2aa2e7a2b283ece8553f3ac/api/src/main/java/edu/cornell/mannlib/vitro/webapp/visualization/utilities/VisualizationCaches.java#L414)", "[personLabels](https://github.com/vivo-project/VIVO/blob/78f042016ea198dae2aa2e7a2b283ece8553f3ac/api/src/main/java/edu/cornell/mannlib/vitro/webapp/visualization/utilities/VisualizationCaches.java#L521)", and "[personToMostSpecificLabel](https://github.com/vivo-project/VIVO/blob/78f042016ea198dae2aa2e7a2b283ece8553f3ac/api/src/main/java/edu/cornell/mannlib/vitro/webapp/visualization/utilities/VisualizationCaches.java#L554)" could benefit from the same improvement... but I would suggest we do that as separate tickets, if at all.

PR #197, opened 2020-10-17
Enable war deployment model

awoods:
Partial resolution to: https://jira.lyrasis.org/browse/VIVO-1443

This update changes VIVO to no longer post the (exploded) war to Tomcat during build, nor create the VIVO_HOME directories.
Instead, the build now produces a war file, `vivo.war` (found in "installer/webapp/target/") that must be "manually" installed by copying it into the Tomcat webapps directory.
When Tomcat/VIVO is started, the VIVO_HOME directory is populated from the standard resources that are now built into the war file.
Other notes:
- Hardcode vivo.all.log file name
- Must now set system property on application deploy: `-Dvivo-dir=/opt/vivo/home/`
- Optionally provide build property to set app-name: `-Dapp-name=vivo`

Once the process for building, deploying and customizing has been tested with this pull-request, I would like to follow-on by simplifying the Maven structure by removing the various "installer" Maven modules.

# How should this be tested?
To test, 
1. Build VIVO and notice that files in VIVO_HOME have not been changed
   - Update your Tomcat system properties (/etc/default/tomcat9 ?) to include the `vivo.dir` system property that references an existing VIVO_HOME directory, such as:
      ```
      JAVA_OPTS="${JAVA_OPTS} -Dvivo-dir=/opt/vivo/home/"
      ```
   - Install VIVO by dropping the generated war file into the Tomcat webapps directory
2. If your VIVO_HOME directory exists but is empty, it will be populated with the default files, including `default.runtime.properties`, `default.applicationSetup.n3` and `default.developer.properties`.
   - If your VIVO_HOME directory exists and is populated, standard files in bin and rdf directories will be updated.
   - If it is desired to keep changes in VIVO_HOME directory you will need to create a checksum digest that is used to determine whether or not to overwrite. Otherwise, the digest is created first deploy if not yet exists.
   
Command to create checksum digest.
```
find /vivo/home -type f | cut -c3- | grep -E '^bin/|^config/|^rdf/' | xargs md5sum > /vivo/home/digest.md5`
```

For testing with a third-tier build, supporting overlay of webapp and home directory.
https://github.com/awoods/vivo-custom-webapp/pull/2

# Interested parties
@VIVO-project/vivo-committers


brianjlowe:
Unless I'm missing something, doesn't the settting of a system property on Tomcat to specify the home directory preclude running more than one VIVO per Tomcat?  This would be a non-starter for me, and, I suspect, for anyone wanting to provide VIVO hosting services (it looked like there was such an individual on the mailing list recently).  Not only do I regularly run multiple VIVOs for development, but this also occurs on clients' servers where there is one (or more) VIVO for testing or data staging, and another hosting the public VIVO site.

awoods:
> Unless I'm missing something, doesn't the settting of a system property on Tomcat to specify the home directory preclude running more than one VIVO per Tomcat? This would be a non-starter for me, and, I suspect, for anyone wanting to provide VIVO hosting services (it looked like there was such an individual on the mailing list recently). Not only do I regularly run multiple VIVOs for development, but this also occurs on clients' servers where there is one (or more) VIVO for testing or data staging, and another hosting the public VIVO site. Edit: ditto the hardcoded log file name.

@brianjlowe : Thank you for raising the concern around deploying multiple VIVOs within a single Tomcat. There are several ways for VIVO to determine its VIVO_HOME directory... I mentioned system properties in this PR because it was the most expedient. The [VitroHomeDirectory.java](https://github.com/vivo-project/Vitro/blob/master/api/src/main/java/edu/cornell/mannlib/vitro/webapp/application/VitroHomeDirectory.java#L79-L81) class defines three mechanisms for providing the VIVO_HOME location to the application:
- [JNDI](https://tomcat.apache.org/tomcat-9.0-doc/jndi-resources-howto.html) - This can be set in each VIVO's `/WEB-INF/web.xml` or `META-INF/context.xml`
- System Properties - Which are described in this PR, but likely only allow for one VIVO deployment per Tomcat
- Build Properties - Which can be set in each VIVO's `/WEB-INF/resources/build.properties`

awoods:
@brianjlowe : as a confirmation, removing the system property, `Dvivo-dir=`, from "/etc/default/tomcat9" and updating the value in "[META-INF/context.xml](https://github.com/vivo-project/VIVO/blob/master/installer/webapp/src/main/webResources/META-INF/context.xml)" allows me to deploy multiple VIVO war files in a single Tomcat. Additionally, the documentation indicates that the `context.xml` file can also live in Tomcat, vs in the war:
https://tomcat.apache.org/tomcat-9.0-doc/config/context.html#Defining_a_context

gneissone:
@awoods Given the above info, what will the recommendation be for those running multiple VIVO instances? Will it be to hardcode the value in context.xml for each fork/instance they plan to run? Or are you suggesting altering this pull request to remove the system property requirement? 

Side comment... I'm not an expert on anti-patterns, but I suspect setting vivo-dir (aka vitro.home) during code compilation is one. That said, I am worried about people struggling to find the right place to define vivo-dir given the many places Tomcat and Java can be installed, and the different folder structures imposed by various package managers. Similar complications if we recommend setting it as a system-wide environment variable. If this is the route we go, we should be meticulous about documenting best practices for setting this variable in common environments.

awoods:
@gneissone / @brianjlowe : Here are some more details regarding the two scenarios:
1. Standard build / deployment, i.e. build VIVO, copy the `war` into the "webapps" directory of Tomcat
   - The location of VIVO_HOME can be provided in any of the following ways:
      - System property provided to Tomcat (as noted above): `-Dvivo-dir=/opt/vivo/home/`, or
      - Provided as parameter at build-time: "mvn clean install `-Dvivo-dir=/opt/vivo/home/", or
      - Provided in `context.xml` within Tomcat: Copy [context.xml](https://github.com/vivo-project/VIVO/blob/master/installer/webapp/src/main/webResources/META-INF/context.xml) into location detailed in [Tomcat documentation](https://tomcat.apache.org/tomcat-9.0-doc/config/context.html#Defining_a_context), changing value of "vivo-dir" in that file
   - In this scenario, the application log file will be named `vivo.all.log`
2. Three tier build, or multiple VIVO webapps being deployed to the same Tomcat
   - See instructions in: https://github.com/awoods/vivo-custom-webapp

gneissone:
@awoods Gave this a spin and in tandem with the Vitro PR and it worked. I have some meandering thoughts and questions.

-I did not have to set an environment variable. It seems like the VIVO home directory is still passed through from the settings.xml file if it's still there? 

-I received an error that the home directory was not empty even though it appeared to be... because OS X likes to add hidden .DS_Store files. Would it make sense to check for the expected folders instead of requiring the folder be entirely empty? Not a big deal either way. The error that was passed to the user on the front end could perhaps be more descriptive. The smoke test says 
` ERROR [StartupStatus] From ApplicationSetup: Failed to initialize the Application.
java.lang.IllegalStateException: '/usr/local/vivo/home/config/applicationSetup.n3' does not exist.`

which is true, but the actual cause of the error was the home directory was not entirely empty so VIVO didn't add the files.

-This seems to force users to change the root user address and default namespace after the first run, then restart VIVO for the changes to take effect. When they login for the second time, they will get warnings for both... and need to manually delete the original root user to make the warnings go away. Would it make sense to pause startup somehow and alert the user to change the root user to their desired email in runtime.properties before continuing startup?

-Last thought, this will mean we can no longer update files in the home directory in our source code, track them with git, rebuild the application and have them overwrite the old files on deployment. For example, updated local ontologies will need to be manually copied into the home directory, whereas in the past upon deployment they would overwrite the old file.

awoods:
Thanks, @gneissone : Did you happen to try: https://github.com/awoods/vivo-custom-webapp ?

gneissone:
Note: The intended way to build this is `mvn clean install` (no settings.xml)

wwelling (approved):

gneissone:
Copying over some discussion that occurred on Slack and adding some responses. Doing my best to wrap my head around the consequences of this change.... though seemingly a straightforward change, a legacy of VIVO quirks makes me ponder how serious these changes (and especially the home folder deployment aspect within Vitro) may be for existing users.

> @gneissone : Thanks for looking at this again. I believe the 3rd bullet is addressed by the follow-on commit in that PR:
https://github.com/vivo-project/VIVO/pull/197/commits/bbc470a2299fce016b32a7c78475aa7241a67a2c
Do you agree?

Yes, requiring the default root user and namespace to be set in installer/webapp/src/main/webResources/META-INF/context.xml prior to build seems like a reasonable solution. I assume the default value of ${vivo-dir} for the home directory is required for if the user sets it via -Dvivo-dir instead of editing context.xml. 

You mentioned a possible 'appName' variable. That seems to have been added to Vitro's context.xml, but not VIVOs. Should it be here, as well, so that we can control the log name? Will it have any effect beyond setting the log name? If not, perhaps the variable should be renamed.

> Regarding the 2nd bullet, we also talked about "merging" of config files as opposed to doing nothing if the VIVO_HOME is not empty. In the meantime, I could update the logic to look for the directories:  - config , and - rdf. 
If they do not exist, overwrite what is in VIVO_HOME with the new start-up files.

> Regarding your "bullet #2":
-I received an error that the home directory was not empty even though it appeared to be... because OS X likes to add hidden .DS_Store files. Would it make sense to check for the expected folders instead of requiring the folder be entirely empty? Not a big deal either way. The error that was passed to the user on the front end could perhaps be more descriptive. The smoke test says
ERROR [StartupStatus] From ApplicationSetup: Failed to initialize the Application. java.lang.IllegalStateException: '/usr/local/vivo/home/config/applicationSetup.n3' does not exist.
..we landed on:
On start-up, files in "home-files.tar" should be copied into VIVO_HOME, whether or not files already exist in that directory
..which is implemented in the PR.
If we should revisit that decision, let's discuss

With this logic, the application will deploy all the contents of home-files.tar every time it starts, even if they have been edited locally. If a user edits 'permission_config.n3' then restarts the application, their changes will be overwritten. I'm not too bothered by this, since I track changes with git and will deploy them with Maven. But I wonder if this is necessary? Does it make sense to have the copy process be part of a 'firsttime' procedure instead of 'everytime' (pardon the use of these particularly loaded terms in VIVO parlance). 

Note, simply copying over the war file makes the application available at localhost:8080/vivo-installer-vivo-1.11.2-SNAPSHOT... which makes sense but isn't very pretty. I suppose the installation instructions can instruct the user to rename the file to whatever they want prior to copying into the webapps directory. This was previously set via the installer file, using the app-name variable.

gneissone:
> Thanks, @gneissone : Did you happen to try: https://github.com/awoods/vivo-custom-webapp ?

Took a look and was able to build with webapp customizations, woohoo! Will it also be possible to overlay home folder customizations? It seems like in its current iteration it pulls the home directory from a snapshot image somewhere.

gneissone (approved):

gneissone:
> > Thanks, @gneissone : Did you happen to try: https://github.com/awoods/vivo-custom-webapp ?
> 
> Took a look and was able to build with webapp customizations, woohoo! Will it also be possible to overlay home folder customizations? It seems like in its current iteration it pulls the home directory from a snapshot image somewhere.

To close the loop on this thread, https://github.com/vivo-community/vivo-custom-webapp has been updated to allow home directory overlays.

wwelling (approved):

roflinn (approved):

awoods:
These instructions do not include mention of updating your "runtime.properties" for enabling i18n languages, but they do explain the updated build and installation process:
https://wiki.lyrasis.org/display/VIVO/VIVO+Build+and+Installation+Process

PR #199, opened 2020-12-04
Update orcidConfirm.ftl

lb-ov:
Hello,
Line 128 have step1 (step1 i18n().orcid_button_step1) i18n key for both cases of the "if" sentence.  For "else" part, it should be Step2 (i18n().orcid_button_step2)


gneissone (approved):
Nice catch, thank you.

**Step one**
Screen Shot 2021-01-05 at 12 16 23 PM

**Step two, before fix**
Screen Shot 2021-01-05 at 12 16 38 PM

**Step two, after fix**
Screen Shot 2021-01-05 at 12 19 37 PM

brianjlowe (approved):

PR #200, opened 2021-01-25
[VIVO-1831] - Remove duplicate text from dropdown menus

gneissone:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1831)**: i18n : In "Create publication" form, dropdown menu has duplicate item

Fixes an issue where both 'select a type' and 'select one' appeared in dropdown menus.

# What's new?
Simply removes the offending text string reference from two template files.

# How should this be tested?
On sprint i18n branch, either click the plus sign next to 'selected publications' or 'editor of' on a person's profile page. Confirm you see both 'select a type' and 'select one' in the publication type dropdown. 

duplicate_item_publication_type_menu_en

Apply patch. Confirm only 'select a type' appears.

Screen Shot 2021-01-25 at 2 35 33 PM

# Additional Notes:
Did not find any other forms that were affected by this, though there certainly are other forms that use the 'select one' logic for the dropdown. The logic appeared when the [buildResourceAndLabelFieldOptions](https://github.com/vivo-project/VIVO/blob/1da890a38e51fcce8d25f1579da414156650cab8/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/configuration/generators/GeneratorUtil.java#L51) method was added to GeneratorUtil.java, but not all form generators use this new method. Simply removing the additional string in the template seems like the easiest solution here, @brianjlowe do you have any thoughts?

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/edit/forms/addEditorshipToPerson.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/edit/forms/addEditorshipToPerson.ftl
@@ -84,7 +84,6 @@ Set this flag on the input acUriReceiver where you would like this behavior to o
         <label for="orgType">${i18n().document_type_capitalized} ${requiredHint}</label>
         <#assign docTypeOpts = editConfiguration.pageData.documentType />
         <select id="typeSelector" name="documentType" acGroupName="document">
-            <option value="" selected="selected">${i18n().select_one}</option>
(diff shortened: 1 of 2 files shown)


awoods (approved):
This looks good and works as expected. I will merge, @brianjlowe , unless you had any comments.

PR #204, opened 2021-01-26
[VIVO-1924] i18n: @en i18n properties are not loaded during first site startup.

dofeldsc:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1924?src=confmacro)**

This PR fixes the bug described in the ticket by rearranging the startup-listeners. 

# How should this be tested?
Test the Bug:
- When starting VIVO for the very first time, i18n strings in English are not loaded. The 'theme' strings are loaded, but the top level all.properties are not. If the i18n cache is disabled, then the strings immediately appear. 
- Start your vivo (with english enabled in runtime.properties) for the first time and visit the main page (language set to en-US), there should be some missing variables (see images in JIRA-Ticket)

Test the fix:
- apply changes
- Start your vivo (with english enabled in runtime.properties) for the first time and visit the main page, everything should look normal and no missing variables

Note: The error does not occur for everyone. For more information see comments in JIRA-Ticket

### DIFF
--- a/api/src/main/java/org/vivoweb/webapp/startup/i18nSetup.java
+++ b/api/src/main/java/org/vivoweb/webapp/startup/i18nSetup.java
@@ -7,6 +7,12 @@ import edu.cornell.mannlib.vitro.webapp.i18n.VitroResourceBundle;
 import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
 
+/**
+  * Configures VIVO to look for i18n properties files prefixed with 'vivo_'.
+  * This listener must be run before any other code that uses resource bundles.
+  * As this listener does not depend on any others, it should be run at or near
+  * the top of the list in startup_listeners.txt.
+  */
 public class i18nSetup implements ServletContextListener {
 
     @Override
--- a/webapp/src/main/webapp/WEB-INF/resources/startup_listeners.txt
+++ b/webapp/src/main/webapp/WEB-INF/resources/startup_listeners.txt
@@ -5,10 +5,8 @@
 #     https://sourceforge.net/apps/mediawiki/vivo/index.php?title=The_StartupManager
 #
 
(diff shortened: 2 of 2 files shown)


brianjlowe (changes requested):
This looks great, Dominik.  A small addition:  would you mind updating this pull request by adding a comment at the top of the org.vivoweb.webapp.startup.i18Setup Java file?  This will help prevent us from inadvertently recreating this bug again the future.  I might suggest wording like:

/**
 * Configures VIVO to look for i18n properties files prefixed with 'vivo_'.
 * This listener must be run before any other code that uses resource bundles.
 * As this listener does not depend on any others, it should be run at or near
 * the top of the list in startup_listeners.txt.   
 */

Edit: we should probably also add a brief one-line version of the comment above the line in startup_listeners.txt indicating that the i18nSetup should remain at the top.

dofeldsc:
Very good idea, I added the comment.

awoods:
@dofeldsc : If I understand @brianjlowe 's suggestion correctly, he is suggesting two comments:
1. The block comment (which you added to [startup_listeners.txt](https://github.com/vivo-project/VIVO/blob/sprint-i18n/webapp/src/main/webapp/WEB-INF/resources/startup_listeners.txt)) should be added to the top of [i18nSetup.java](https://github.com/vivo-project/VIVO/blob/sprint-i18n/webapp/src/main/webapp/WEB-INF/resources/startup_listeners.txt)
2. A one-line comment to be added to [startup_listeners.txt](https://github.com/vivo-project/VIVO/blob/sprint-i18n/webapp/src/main/webapp/WEB-INF/resources/startup_listeners.txt)

dofeldsc:
yes you are right, i will change it. Sorry. I should not do several things at the same time...

awoods (approved):

brianjlowe (approved):

PR #211, opened 2021-02-02
Escaping single quotes as ascii codes in publication count .ftl

nicalico:
Resolves https://jira.lyrasis.org/browse/VIVO-1814

### What does this pull request do?
It is a known issue that some properties containing single quotes can cause conflict when Javascript is inserted into ftl files, hence the need to use the `js_string` builtin to escape problematic characters (see https://jira.lyrasis.org/browse/VIVO-1842). 

In the two files modified in this PR, however, the conflict occurs in the html written by the Javascript, where the single quotes are displayed in the attributes of an `img` tag.

> image

In [Freemarker >2.3.24](https://freemarker.apache.org/docs/ref_builtins_string.html#ref_builtin_html), those special characters are automatically escaped as ascii codes. However, if VIVO is installed with an older version of Freemarker, it would have to be escaped with the (deprecated) `html` builtin.

This PR proposes to fix the bug by using the `replace` builtin, in order to manually convert the single quotes into ascii code. While not very elegant, this method has the merit of working regardless of the version of Freemarker used.

### How should this be tested?
With the fr_CA context selected, browse to the profile of an individual, then check the mouseover text of the Information icon in the left pane.
> image

The text should read as :  **"Ces chiffres sont basés uniquement sur les publications qui ont été chargées dans cette application VIVO. Si c'est votre profil, vous pouvez entrer d'autres publications ci-dessous."**

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/visualization/publication/personPublicationCountNoSparkline.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/visualization/publication/personPublicationCountNoSparkline.ftl
@@ -59,7 +59,7 @@
 
                 var td1Text = totalPublicationCount;
                 var td2Text = "";
-                var infoImgText = "<img class='infoIcon' src='" + infoIconSrc + "' height='14px' width='14px' alt='${i18n().info_icon}' title='${i18n().numbers_based_on_publications_in_vivo}' />";
+                var infoImgText = "<img class='infoIcon' src='" + infoIconSrc + "' height='14px' width='14px' alt='${i18n().info_icon?replace("'","&#39;")}' title='${i18n().numbers_based_on_publications_in_vivo?replace("'","&#39;")}' />";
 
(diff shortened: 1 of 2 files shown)


awoods:
The code changes here look good. However, when I try to test the functionality, I am not getting the sparkline (and associated info link) with i18n languages enabled. On the other hand, if I disable i18n and only have the default `en` enabled, I do see the sparkline and info link... but not the `fr_CA` info link.
@nicalico : can you detail how you are testing this.

nicalico:
I have indeed noticed problems with the sparkline: for instance, when refreshing the page, it just disappears, and the only way to get it back seems to browse out then back in the individual profile (which makes testing rather frustrating).

I wondered whether it was a local Tomcat caching issue, but there's obviously another (unrelated) bug in here.

awoods (approved):
Works as advertised.

PR #214, opened 2021-02-04
[VIVO-1930] - Update example.runtime.properties

gneissone:
See https://github.com/vivo-project/VIVO/pull/194 for details
https://jira.lyrasis.org/browse/VIVO-1930

PR #222, opened 2021-02-11
Update example.runtime.properties

awoods:
See #194 for details

Resolves: https://jira.lyrasis.org/browse/VIVO-1930

### DIFF
--- a/home/src/main/resources/config/default.runtime.properties
+++ b/home/src/main/resources/config/default.runtime.properties
@@ -52,11 +52,8 @@ argon2.time = 1000
   # the "Contact Us" form will be disabled and users will not be notified of
   # changes to their accounts.
   #   Example:
-  #     email.smtpHost = smtp.mydomain.edu
-  #     email.replyTo = vivoAdmin@mydomain.edu
-  #
-email.smtpHost =
-email.replyTo =
+# email.smtpHost = smtp.mydomain.edu
+# email.replyTo = vivoAdmin@mydomain.edu
 
   #
   # URL of Solr context used in local VIVO search. This will usually consist of:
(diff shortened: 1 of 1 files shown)


awoods:
Supersedes: https://github.com/vivo-project/VIVO/pull/214

brianjlowe:
Looks great.  My only question is whether it would be more consistent to comment out the (empty) SMTP/replyTo properties rather than leaving them uncommented with empty values.

gneissone:
> Looks great. My only question is whether it would be more consistent to comment out the (empty) SMTP/replyTo properties rather than leaving them uncommented with empty values.

If I recall correctly VIVO complains if the properties are missing. Let's see if my memory is correct...

brianjlowe (dismissed):

gneissone:
@awoods Just occurred to me we should update this value, too https://github.com/vivo-project/VIVO/blob/52139befe81c844aabc66007b26bb0c21e5485ea/home/src/main/resources/config/default.runtime.properties#L191

es_GO doesn't exist for VIVO anymore, correct? I imagine the logic for using the non-standard locale is still in the source, but the language files have been renamed to just 'es' if I recall correctly.

awoods:
> @awoods Just occurred to me we should update this value, too
> 
> https://github.com/vivo-project/VIVO/blob/52139befe81c844aabc66007b26bb0c21e5485ea/home/src/main/resources/config/default.runtime.properties#L191
> 
> es_GO doesn't exist for VIVO anymore, correct? I imagine the logic for using the non-standard locale is still in the source, but the language files have been renamed to just 'es' if I recall correctly.

Since that will require additional testing, let's keep the scope of this set of PRs constrained... and revisit `es_GO` as a follow-on activity.

brianjlowe (approved):

awoods:
@gneissone : since both you and I have fingerprints on this PR, it would be nice if someone else performed the "squash/merge".

PR #228, opened 2021-03-09
revert deployment changes

wwelling:
Revert to previous deployment strategy with maintaining Docker build and GitHub workflows.

Requires https://github.com/vivo-project/Vitro/pull/217


brianjlowe:
So all seems well doing the normal Maven install.  But if I run docker-compose up, I get a functioning VIVO and Solr but all of the i18n properties on the page are just "ERROR: Text bundle 'all' not found" until I set my browser back to en-US.  It seems to behave as if it's using a version of Vitro that is missing commit 6b612316eedf2c1783a1c4b85c08be99cea1b22a (use fallback locale...).  Is there something else I'm missing here?

wwelling:
@brianjlowe did you checkout out the Vitro revert PR? https://github.com/vivo-project/Vitro/tree/revert-deployment-strategy

brianjlowe:
OK, I now have gotten this to work.  My first problem (above) was due to my not knowing how to rebuild the image.  William suggested `docker container prune`, `docker image prune` and `docker-compose build`.  For good measure, I'm now running all three of those in order in between `docker-compose down` and `docker-compose up` to make sure everything gets updated.  Seems fine.  My second problem was that it was still picking up an old runtime.properties file without vitro.defaultNamespace and throwing a fatal error on the smoke tests.  So far as I can tell, this is because I still had an old vivo-home directory from the previous build before applying these pull requests, and the runtime.properties from there was being used instead of the one from installer/home/target/vivo/config.  I'm not sure if this is expected behavior or if it suggests that something in startup.sh should change so as not to try to use vivo-home.

nicalico:
**With those branches checked out:**

- VIVO: upstream/revert-deployment-strategy
- Vitro : upstream/revert-deployment-strategy
- VIVO-languages : upstream/main
- Vitro-languages : upstream/main

**Built with the following command:**

mvn -T 1.5C -s settings.xml -DskipTests=true -U clean install

**In the following contexts:**

- On a CentOS 7 VM, in vanilla mode
- On a CentOS 7 VM, with UQAM overlays
- In Eclipse IDE, with UQAM overlays

**Conclusion:** appears to work fine in all those cases.

litvinovg:
Tested on CentOS 7 with Tomcat 8.5.55. Works fine.
Used those branches:
    VIVO: revert-deployment-strategy
    Vitro : revert-deployment-strategy
    VIVO-languages : main
    Vitro-languages : main

brianjlowe (approved):

brianjlowe:
@nicalico  or @litvinovg Thanks for the testing: if one of you is able to add your approval checkmark to the PR, I'll go ahead and merge.

litvinovg (approved):
Looks good to me.

PR #231, opened 2021-03-26
Retrieve sitemap profile URLs only after consuming result set in orde…

brianjlowe:
…r to avoid transaction errors with TDB. Resolve https://jira.lyrasis.org/browse/VIVO-1910

**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1910)**: 

Fixes bug where sitemap.xml failed to generate when using TDB.

# What's new?
Profile URLs for the sitemap are retrieved after iterating through the ResultSet of person URIs.  This avoids the problem of trying to start a new TDB transaction while inside an existing one.  The sitemap will also retrieve a maximum of 50000 to conform to Google's limit.

# How should this be tested?
Use with TDB (default) and create a person or two from SiteAdmin if your database doesn't already have some.  Without the PR, a request for /sitemap.xml will result in an error in the browser ("no elements found ...").  After applying the pull request, sitemap.xml will render with a list of person URLs.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/org/vivoweb/webapp/sitemap/SiteMapServlet.java
+++ b/api/src/main/java/org/vivoweb/webapp/sitemap/SiteMapServlet.java
@@ -2,29 +2,40 @@
 
 package org.vivoweb.webapp.sitemap;
 
-import org.apache.commons.lang3.StringUtils;
-import org.apache.jena.query.QuerySolution;
-import edu.cornell.mannlib.vitro.webapp.controller.VitroHttpServlet;
-import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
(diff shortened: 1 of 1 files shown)


gneissone (approved):
Tested and it works! Looks like the 50k limit is for individual sitemap documents, though it is possible to reference multiple sitemap documents from a sitemap index. So we could in theory support > 50k people in a VIVO if we wanted to make a followup task.

brianjlowe:
Thanks, Benjamin.  Good idea about the multiple sitemap documents.  After doing this PR, it also occurred to me that it might be a nice improvement to retrieve the sitemap entries from the search index instead of from the triple store.  That would avoid submitting profiles for "external" persons who might be hidden from the search index but still exist as foaf:Persons.  Possibly something to chat about on tomorrow's call.

Issue #3610, opened 2021-12-20
1.12.1 not building 


gneissone:
Installing 1.12.1 does not seem to work correctly. Languages don't seem to be included, so if you manage to build the application, VIVO complains about not finding the i18n directory upon startup. 

**Additional context**
Behavior was replicated by @chenejac and @brianjlowe

fmdelvalle:
I could work around this issue by:
1) copying the VIVO/installer directory outside
2) Copying the language files to the installer (these are for Spanish, and assumes the installer is in 'installer' folder, replace as you need):
cp -rf VIVO-languages/es/home/src/main/resources/rdf/i18n/es/* installer/home/src/main/resources/rdf/ &&
cp -rf Vitro-languages/es/webapp/src/main/webapp/* installer/webapp/src/main/webapp/ &&
cp -rf VIVO-languages/es/webapp/src/main/webapp/* installer/webapp/src/main/webapp/ &&
3) building with a custom installer: mvn install -s settings.xml -Dvivo-installer-dir=<path to copied installer folder>

With these I'm able to start the app.

PR #3611, opened 2021-12-20
Uncomment missing language dependencies

gneissone:
**[Issue](https://github.com/vivo-project/VIVO/issues/3610)**: resolves https://github.com/vivo-project/VIVO/issues/3610

Re-adds language dependencies into pom.xml files. VIVO and Vitro languages must now always been included, even when using VIVO with a single language.

# How should this be tested?
Follow instructions on how to install VIVO at https://wiki.lyrasis.org/display/VIVODOC112x/Installing+VIVO. If VIVO installs and does not complain about missing language files or strings upon startup, it was a success. 

# Interested parties
@roflinn

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ENV JAVA_OPTS="${JAVA_OPTS} -Dtdb:fileMode=$TDB_FILE_MODE"
 RUN mkdir /usr/local/vivo
 RUN mkdir /usr/local/vivo/home
 
-COPY ./installer/home/target/vivo-installer-home-1.12.1 /vivo-home
+COPY ./installer/home/target/vivo /vivo-home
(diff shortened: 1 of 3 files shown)


litvinovg (approved):
Works good.

wwelling (approved):

roflinn (changes requested):
Can we change the need in the Dockerfile to need the version number hardcoded?

Ideally update the logic that has the directory named with the version number.

gneissone:
@roflinn I can pull that change out since it's really a separate issue. Something has changed in the pom files since the Dockerfile was created. Not sure what, but the Dockerfile broke due to the change in target directory name.

gneissone:
@roflinn found the change that was breaking the Dockerfile and re-added a line in installer/home/pom.xml that was removed: `<finalName>${app-name}</finalName>`

PR #3613, opened 2021-12-28
Fallback in case of missing most specific type for microformat macro

### LINKED ISSUE #3612: Freemarker error thrown related to microformats
**Describe the bug**
This bug was reported way back in 2018 here: https://groups.google.com/g/vivo-tech/c/PEIjS7kuIeY/m/odmwhKXeBwAJ

**To Reproduce**
Unfortunately I have no idea why this started happening for me in one of my projects. I loaded a large file and it suddenly started happening for a specific class (vivo:Department) but I perhaps it has something to do with retrieving the most specific type. I couldn't see anything in the triples that would suggest an issue (mostSpecificType was correctly in the kb-inf graph). 

**Expected behavior**
The page the load without issue. 

**Screenshots**
Full error and stack trace was same as in vivo-tech issue posted above. 
```
Caused by: FreeMarker template error: When calling macro "getItemType", required parameter "type" (parameter #1) was specified, but had null/missing value. ---- Tip: If the parameter value expression on the caller side is known to be legally null/missing, you may want to specify a default value for it with the "!" operator, like paramValue!defaultValue. ---- ---- FTL stack trace ("~" means nesting-related): -

Failed at: @getItemType mst, "organization" [in template "lib-microformats.ftl" in macro "sectionSchema" at line 10, column 9]
- Reached through: @mf.sectionSchema individual [in template "individual-vitro.ftl" at line 19, column 60]
- Reached through: #include "individual-vitro.ftl" [in template "individual.ftl" at line 32, column 1]
- Reached through: #include "individual.ftl" [in template "individual--foaf-organization.ftl" at line 31, column 1]
```

**Environment (please complete the following information):**
Linux, Chrome, Tomcat 8, VIVO 1.12.x, Solr

gneissone:
**[JIRA Issue](https://github.com/vivo-project/VIVO/issues/3612)**: Fixes #3612 

https://groups.google.com/g/vivo-tech/c/PEIjS7kuIeY/m/odmwhKXeBwAJ

Prevents a Freemarker error encountered in rare cases.

# What's new?
Adds blank fallback in case of issue retrieving a most specific type value within the lib-microformats.ftl file.

# How should this be tested?
Search an individual's page source for a microformat value, e.g. `itemscope itemtype="http://schema.org/Thing"` before applying change. After applying change, you should still see the microformat value on the page... but it should also prevent a freemarker error from being thrown in case the macro was unable to find a most specific type (for whatever reason...). 

# Additional Notes:
Unfortunately I don't know exactly why this error is happening... this is more of a band-aid than an actual fix for the core problem, but it doesn't hurt to work around the problem in lieu of investing more time finding the root cause.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/lib/lib-microformats.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/lib/lib-microformats.ftl
@@ -11,12 +11,11 @@
     <#elseif individual.event()>
         itemscope itemtype="http://schema.org/Event"
 	<#elseif individual.infoContentEntity()>
-	   <#assign mst = "Document" />
 		<#list individual.mostSpecificTypes as type >
 			<#assign mst = type />
 			<#break>
 		</#list>
-       <@getItemType mst!"" "infoContentEntity"/>
+       <@getItemType mst!"Document" "infoContentEntity"/>
     <#else>
         itemscope itemtype="http://schema.org/Thing"
     </#if>
--- a/webapp/src/main/webapp/templates/freemarker/lib/lib-microformats.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/lib/lib-microformats.ftl
@@ -7,7 +7,7 @@
 			<#assign mst = type />
 			<#break>
 		</#list>
-        <@getItemType mst "organization"/>
+        <@getItemType mst!"" "organization"/>
     <#elseif individual.event()>
         itemscope itemtype="http://schema.org/Event"
 	<#elseif individual.infoContentEntity()>
@@ -16,7 +16,7 @@
 			<#assign mst = type />
 			<#break>
 		</#list>
-    	<@getItemType mst "infoContentEntity"/>
+       <@getItemType mst!"" "infoContentEntity"/>
     <#else>
         itemscope itemtype="http://schema.org/Thing"
     </#if>


webapp/src/main/webapp/templates/freemarker/lib/lib-microformats.ftl
```diff
@@ -16,7 +16,7 @@
 			<#assign mst = type />
 			<#break>
 		</#list>
-    	<@getItemType mst "infoContentEntity"/>
+       <@getItemType mst!"" "infoContentEntity"/>
```
chenejac, line 19: Probably this should be mst!"Document", and line 14 in this file should be removed.

chenejac (changes requested):
@gneissone thanks for this patch. I am personally surprised that the problem doesn't appear in other files when mostSpecificTypes is not calculated well for any reason. For instance, in the line https://github.com/vivo-project/VIVO/blob/main/webapp/src/main/webapp/templates/freemarker/body/individual/individual--foaf-organization.ftl#L20. Although I would prefer that we find the root of the problem, I would approve this contribution after addressing the small issue I have commented (see my comment for your code).

webapp/src/main/webapp/templates/freemarker/lib/lib-microformats.ftl
gneissone: Thanks, I have made the change.

chenejac (approved):

litvinovg (approved):

PR #3776, opened 2022-10-06
[ i18n sprint ] Fixes for new individual form 

litvinovg:
Eliminates the need to combine translations in new individual form

# Interested parties
@chenejac


chenejac (approved):
Thanks. Works as it is described.

wwelling (approved):

PR #3784, opened 2022-10-19
[i18n sprint] Language-neutral help text in autocomplete forms

litvinovg:
[Companion Vitro-languages PR](https://github.com/vivo-project/Vitro-languages/pull/64)
[Companion Vitro PR](https://github.com/vivo-project/Vitro/pull/338)
Use language-neutral selectAnExistingOrCreateNewOne translation key as help text in forms with auto-completion

# How should this be tested?
Try out forms with auto-completion to see new text

# Interested parties
@tawahle

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/edit/forms/autoCompleteDataPropForm.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/edit/forms/autoCompleteDataPropForm.ftl
@@ -72,7 +72,7 @@ Also multiple types parameter set to true only if more than one type returned-->
     };
     var i18nStrings = {
         selectExisting: '${i18n().select_an_existing?js_string}',
-        orCreateNewOne: '${i18n().or_create_new_one?js_string}',
+        selectAnExistingOrCreateNewOne: '${i18n().select_an_existing_or_create_a_new_one?js_string}',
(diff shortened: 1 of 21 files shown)


webapp/src/main/webapp/templates/freemarker/edit/forms/autoCompleteDataPropForm.ftl
```diff
@@ -72,7 +72,7 @@ Also multiple types parameter set to true only if more than one type returned-->
     };
     var i18nStrings = {
         selectExisting: '${i18n().select_an_existing?js_string}',
```
chenejac, line 74: Do we need selectExisting field now? It looks as the new field is only needed? Also, if the selectExisting is needed, please rename it to selectAnExisting due to consistence.

chenejac (changes requested):

webapp/src/main/webapp/templates/freemarker/edit/forms/autoCompleteDataPropForm.ftl
litvinovg: You are right, we don't need it. I removed it.

chenejac (approved):

wwelling (approved):

PR #3808, opened 2023-01-13
Use fallbacks for concept labels with en-US, en, and without language…

litvinovg:
If concept doesn't have label with language tag it can't be used in capability map.
This is a fix for capability map regression from i18n sprint.

# What's new?
Use fallbacks for concept labels with en-US, en, and without language tag in visualization caches.

# How should this be tested?
- Use a concept with label without language tag.
- Search for concept in capability map.
- Concept should be shown on the map.

# Interested parties
@chenejac @bkampe


chenejac (approved):
@litvinovg thanks

Issue #3528, opened 2020-11-18
VIVO-1942: jQuery version flagged as security vulnerability 


chenejac:
[Benjamin Gross](https://github.com/gneissone) (Migrated from [VIVO-1942](https://vivo-project.atlassian.net/browse/VIVO-1942)) said:

A client's security audit software has identified the version of jQuery included with VIVO (1.12.4) as a security vulnerability of 'medium' severity. 

jQuery v 1.12.4 was released in May 2016. The v1 line hasn't been updated since then. I don't have any details on what the potential issues are with using the library, but it seems prudent to update to a modern version if we plan to keep using the library.

PR #3921, opened 2023-11-29
Fixed jquery selector for capability map tabs

milospp:
Fixes the unresponsive behavior of capability map nodes in the Nemo theme when clicked. The issue stemmed from an incorrect jQuery selector used in the theme's jQuery version.

# What's new?
Resolved the jQuery selector issue by adding single quotes to the value of the href attribute:
    $(".titles li a[href='#" + name + "']")...
    

# How should this be tested?
1. Open the Capability Map tab.
2. Search for an item.
3. Click on a node to verify if it opens the information tab on the right side.
4. Test this across all themes for thorough validation.

image

### DIFF
--- a/webapp/src/main/webapp/js/visualization/capabilitymap/graph_new.js
+++ b/webapp/src/main/webapp/js/visualization/capabilitymap/graph_new.js
@@ -388,7 +388,7 @@ FullResultQueryUnit.prototype.fetch = function() {
 
 var showPanel = function(name) {
     $(".titles li").removeClass("activeTab");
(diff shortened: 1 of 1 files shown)


gneissone:
I know it's not the point of this issue and PR and it's a non-trivial task, but would be great if we could upgrade the jQuery version. The version VIVO uses is from 2016 and is flagged as a security risk. https://github.com/vivo-project/VIVO/issues/3528

chenejac (approved):
@milospp well done. Thanks

PR #3927, opened 2023-12-14
Fix for dates document modifier

litvinovg:
In case data contain multiple ranges it prevents document modifier from sending multiple ranges to the same search field, which doesn't support multiple ranges.

# Interested parties
@VIVO-project/vivo-committers


chenejac (approved):
@litvinovg thanks for this fix

PR #3928, opened 2023-12-15
Translations for search related individuals

litvinovg:
[Related Vitro PR](https://github.com/vivo-project/Vitro/pull/431)

Translations for VIVO individuals presented on search results page

@chenejac Added translations for different languages. We need reviews from native speakers.

### DIFF
--- a/home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vivo.n3
+++ b/home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vivo.n3
@@ -1,0 +1,11 @@
+@prefix : <https://vivoweb.org/ontology/vitro-search-individual/> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+
+:filter_publication_year rdfs:label "Erscheinungsjahr"@de-DE .
+:field_persons rdfs:label "Personen"@de-DE .
+:range_filter_dates rdfs:label "Dates"@de-DE .
+:filter_organizations rdfs:label "Organisationen"@de-DE .
+:filter_persons rdfs:label "Personen"@de-DE .
+:field_dates rdfs:label "Dates"@de-DE .
+:field_organizations rdfs:label "Organisationen"@de-DE .
+:field_publication_year rdfs:label "Erscheinungsjahr"@de-DE .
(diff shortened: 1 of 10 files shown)


ivanmrsulja (changes requested):

chenejac (changes requested):
@litvinovg the search query (created at the home page) is lost if locale is changed. I am wondering whether you can switch http method of the form to be GET instead of POST at the home page? I mean something similar as you already did for Vitro [here](https://github.com/vivo-project/Vitro/pull/382/files#diff-d61c7d638a762d87325baba90af65244adf6974987308ebfbc57a7ba7d014a86L37)? It should be also applied for VIVO themes (I suggest to do that for all three themes):

- https://github.com/vivo-project/VIVO/blob/75a1df02cb30b4c3d7cf2c571ee9b925fce213b1/webapp/src/main/webapp/themes/wilma/templates/page-home.ftl#L47
- https://github.com/vivo-project/VIVO/blob/75a1df02cb30b4c3d7cf2c571ee9b925fce213b1/webapp/src/main/webapp/themes/tenderfoot/templates/page/page-home.ftl#L43
- https://github.com/vivo-project/VIVO/blob/75a1df02cb30b4c3d7cf2c571ee9b925fce213b1/webapp/src/main/webapp/themes/nemo/templates/page-home.ftl#L43

litvinovg:
> @litvinovg the search query (created at the home page) is lost if locale is changed. I am wondering whether you can switch http method of the form to be GET instead of POST at the home page? I mean something similar as you already did for Vitro [here](https://github.com/vivo-project/Vitro/pull/382/files#diff-d61c7d638a762d87325baba90af65244adf6974987308ebfbc57a7ba7d014a86L37)? It should be also applied for VIVO themes (I suggest to do that for all three themes):
> 
>     * https://github.com/vivo-project/VIVO/blob/75a1df02cb30b4c3d7cf2c571ee9b925fce213b1/webapp/src/main/webapp/themes/wilma/templates/page-home.ftl#L47
> 
>     * https://github.com/vivo-project/VIVO/blob/75a1df02cb30b4c3d7cf2c571ee9b925fce213b1/webapp/src/main/webapp/themes/tenderfoot/templates/page/page-home.ftl#L43
> 
>     * https://github.com/vivo-project/VIVO/blob/75a1df02cb30b4c3d7cf2c571ee9b925fce213b1/webapp/src/main/webapp/themes/nemo/templates/page-home.ftl#L43

Done

chenejac (approved):
Well done

milospp (approved):

Issue #3939, opened 2024-01-30
Add bind address option to  .env docker-compose


isl3:
As far as I can tell the docker compose will always bind to 0.0.0.0:8983 and 0.0.0.0:8080 respectfully, this causes me trouble as I am trying to run vivo behind a reverse proxy and not expose solr to the entire network.

Please the option to configure what address the containers or the network binds to on the host system.

My current work around is to edit the docker compose before running `docker compose up -d` as follows.
```
  solr:
    image: vivoweb/vivo-solr:latest
    environment:
      - RESET_CORE=${RESET_CORE}
      - VERBOSE=${VERBOSE}
    ports:
      - "127.0.0.1:8983:8983"
    networks:
      - vivo

  tomcat:
   .....
    ports:
      - "127.0.0.1:8080:8080"
```

chenejac:
@isl3 thanks for reporting this. @wwelling created a [PR](https://github.com/vivo-project/VIVO/pull/3940) which should address your issue. It would be great if you can test the solution provided by William and report us back if that is in accordance with your expectations and needs.

isl3:
I've tested the PR and it works as I would expect it to, thank you for the quick work. 
It might be helpful to some there was a comment somewhere saying VIVO_HOST_PORT/SOLR_HOST_PORT can also specify host IP, but this should be obvious to anyone who knows how docker-compose files work so not an large concern.

chenejac:
@isl3 thanks for confirmation. Can you please review [PR](https://github.com/vivo-project/VIVO/pull/3940) and approve it in the case it is working for you? Moreover, you can post your comment there. If @wwelling finds that reasonable he might consider to include that in the PR. Thank you in advance.

PR #3940, opened 2024-02-06
Improved docker setup

wwelling:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**: https://github.com/vivo-project/VIVO/issues/3939

This PR provides more configurability for docker compose and build of the VIVO docker image. Primarily addressing port mapping configuration via `.env` file for `docker-compose.yml` variable substitution.

# What's new?

1. solr and tomcat service port mapping configurable via `.env`
2. ability to specify VIVO home path within the container
3. local mount for VIVO solr core data to persist through container restarts

# How should this be tested?

This can be tested by following https://github.com/wwelling/VIVO/blob/issue-3939/README.md#docker-compose. For more exhaustive testing the `.env` file can be modified between container restarts for desired effect.

# Interested parties
@chenejac, @isl3

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,22 @@
-LOCAL_VIVO_HOME=./vivo-home
-RESET_HOME=false
-RESET_CORE=false
-VERBOSE=no
+# Variable substitution for docker-compose.yml
+
+# solr service variable substitution
+SOLR_RESET_CORE=false
+SOLR_VERBOSE=no
+
+SOLR_HOST_PORT=8983
+SOLR_CONTAINER_PORT=8983
+
+SOLR_CORES=./vivo-cores
+
+# tomcat service variable substitution
+VIVO_RESET_HOME=false
+VIVO_VERBOSE=no
+
+VIVO_TDB_FILE_MODE=direct
+
+VIVO_HOST_VIVO_HOME=./vivo-home
+VIVO_CONTAINER_VIVO_HOME=/usr/local/vivo/home
+
+VIVO_HOST_PORT=8080
+VIVO_CONTAINER_PORT=8080
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,10 +5,12 @@ services:
   solr:
     image: vivoweb/vivo-solr:latest
     environment:
-      - RESET_CORE=${RESET_CORE}
-      - VERBOSE=${VERBOSE}
+      - RESET_CORE=${SOLR_RESET_CORE}
+      - VERBOSE=${SOLR_VERBOSE}
     ports:
-      - 8983:8983
+      - ${SOLR_HOST_PORT}:${SOLR_CONTAINER_PORT}
+    volumes:
+      - ${SOLR_CORES}:/opt/solr/server/solr/mycores
     networks:
       - vivo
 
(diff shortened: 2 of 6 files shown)


.env
```diff
@@ -1,4 +1,22 @@
-LOCAL_VIVO_HOME=./vivo-home
-RESET_HOME=false
-RESET_CORE=false
-VERBOSE=no
+# Variable substitution for docker-compose.yml
```
ivanmrsulja, line 1: What do you think about separating the .env file into 2 files:
- .env.solr
- .env.tomcat

Would that make it easier to configure for the end user?

docker-compose.yml
```diff
@@ -19,18 +21,26 @@ services:
...
-      - 8080:8080
+      - ${VIVO_HOST_PORT}:${VIVO_CONTAINER_PORT}
     volumes:
-      - ${LOCAL_VIVO_HOME}:/usr/local/vivo/home
+      - ${VIVO_HOST_VIVO_HOME}:${VIVO_CONTAINER_VIVO_HOME}
     networks:
       - vivo
+    depends_on:
+      - solr
+    entrypoint:
```
ivanmrsulja, line 38: Is it a better practice to add a healthcheck as described here (https://medium.com/@nikjos/docker-compose-health-check-for-apache-solr-251c04775e24) to Solr container, and then just say that the depends_on condition for tomcat container is "service_healthy" Solr container? IMO, that would remove the need for custom check script and improve potential integration with orchestration tools if somebody decides to use one.

ivanmrsulja (commented):
LGTM and works as intended! I left some comments that are open for discussion, if you have time, so please check them out :slightly_smiling_face:

isl3 (approved):

isl3:
It might be helpful to some there was a comment somewhere saying VIVO_HOST_PORT/SOLR_HOST_PORT can also specify host IP, but this should be obvious to anyone who knows how docker-compose files work so not an large concern.

.env
wwelling: The `.env` currently is for substituting variables for the `docker-compose.yml` and done without additional docker compose command arguments. These variables are technically not service environment variables unless used to set under `environment` of each service. If necessary, an environment file per service can be added and used for the environment variables of the service. At this time, I opted against env file per service as there is more build time variables (build args, volume, and port variables) needed than actual service environment variables.

docker-compose.yml
wwelling: Yes absolutely. Shortly after I put in this PR I did exactly that with another project with docker compose.

Would this be a reasonable health check for Solr?
```
    healthcheck:
      test: ["CMD-SHELL", "curl -Isf http://localhost:8983/solr/vivocore/admin/ping || exit 1"]
      interval: 15s
      timeout: 5s
      retries: 3
...
    depends_on:
      solr:
        condition: service_healthy
```

wwelling: FYI, I read that same article but had difficulty with the health check command they used. I will test a bit before adding additional commit.

ivanmrsulja: > Yes absolutely. Shortly after I put in this PR I did exactly that for another project with docker compose.
> 
> Would this be a reasonable health check for Solr?
> 
> ```
>     healthcheck:
>       test: ["CMD-SHELL", "curl -Isf http://localhost:8983/solr/vivocore/admin/ping || exit 1"]
>       interval: 15s
>       timeout: 5s
>       retries: 3
> ...
>     depends_on:
>       solr:
>         condition: service_healthy
> ```

This seems like a good approach, nice work.

ivanmrsulja (approved):
Looks excellent now 🙂 great work!

.env
ivanmrsulja: I understand, then you should leave it as is. I was just wondering what's your opinion and did my suggestion cross your mind at some point during development 😄.

chenejac (approved):
@wwelling well done. Thanks for this. I don't have any complaint, meaning from my point of view this is ready to be merged.

litvinovg:
@chenejac @ivanmrsulja Can we write down more information about what was tested and reviewed?

chenejac:
> @chenejac @ivanmrsulja Can we write down more information about what was tested and reviewed?

Tested by Ian Slatter, tested and reviewed by Ivan Mrsullja (linux), reviewed by Dragan Ivanovic.

Issue #3951, opened 2024-03-11
Query text parameter not sanitized 


gneissone:
**Describe the bug**
This got flagged by a security scanning application as a potential vulnerability. Pagination in search results is handled via URL parameters and javascript. The input is not sanitized, however, so a bad actor could execute something via the VIVO site's domain.

**To Reproduce**
For example, try this path on any VIVO running the latest code:
{vivo url}/search?querytext=</script><script>alert("uh%20oh");</script>

**Expected behavior**
The arbitrary javascript passed via the URL should not be executed

**Screenshots**
![Screenshot 2024-03-11 at 12 53 57](https://github.com/vivo-project/VIVO/assets/10748475/e5da7478-21bc-422c-a2d3-46f4b7349140)

**Environment (please complete the following information):**
 - Browser: Chrome
 - Tomcat version: 9.0.78
 - VIVO version: 1.14.1-SNAPSHOT
 - Apache Solr 9.3.0

**Additional context**
https://github.com/vivo-project/Vitro/blob/03517df59ab02108f81f19d8ff383e20f9c556ca/webapp/src/main/webapp/templates/freemarker/body/search/search-pagedResults.ftl#L55

PR #3952, opened 2024-03-12
Sanitized search-related freemarker variables

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3951)**

[Vitro PR](https://github.com/vivo-project/Vitro/pull/447)

Sanitized Freemarker variables in search templates.

# How should this be tested?
* Reproduce the problem in the issue
* Test that the pull request fixes the issue
* Test that search field is working in each theme

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/webapp/src/main/webapp/themes/nemo/templates/search.ftl
+++ b/webapp/src/main/webapp/themes/nemo/templates/search.ftl
@@ -8,7 +8,7 @@
 
         <form id="search-form" action="${urls.search}" name="search" role="search" accept-charset="UTF-8" method="GET"> 
             <div id="search-field">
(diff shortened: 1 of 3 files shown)


chenejac (approved):
@litvinovg looks nice.

I have tested all themes and it works as described with the exception of the nemo theme. For the nemo theme after selection of a filter, the initial query is not anymore in the url path, and therefore results are not as expected, only filter is taken into account, but not search query. I think this issue is not introduced by this PR (therefore I will approve this PR), but please check whether you can fix that.

P.S. I didn't check to use html tags in the n3 configuration of the extended search

Issue #3987, opened 2024-05-27
Image uploads


chenejac:
**Describe the bug**
Image uploads is broken for non administrators. Reported by @gneissone 

> Anyone having trouble with image uploads with recent versions of VIVO? I'm getting these errors in the logs:
2024-05-13 19:09:33,509 ERROR [FileInfo] File had no filename: 'https://researchers.mydomain.edu/individual/n23090'
2024-05-13 19:09:33,510 INFO [FileServingServlet] Failed to serve the file at '/file/n50145/thumbnail_totoro.jpg' -- edu.cornell.mannlib.vitro.webapp.filestorage.serving.FileServingServlet$FileServingException: The request path is not valid for the File servlet: '/file/n50145/thumbnail_totoro.jpg'
Images work if I'm logged in as an admin and do a hard refresh, otherwise they appear broken.

**To Reproduce**
1. Run VIVO
2. Log in as a self editor
3. Upload an image for a profile
4. Check logs and see error

PR #3988, opened 2024-05-29
Added permissions required to upload images

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3987)**

[Vitro PR](https://github.com/vivo-project/Vitro/pull/468)

Added permissions required to upload main images.

# How should this be tested?
* Reproduce the issue
* Test that users with following roles able to upload/remove images: Admin, Curator, Editor, Self editor

# Interested parties
@VIVO-project/vivo-committers

# Reviewers' expertise

1. Access control

# Reviewers' report template
## General comment
A reviewer should provide here comments and suggestions for requested changes if any.
## Testing
A reviewer should briefly describe here how it was tested
## Code reviewing
A reviewer should briefly describe here which part was code reviewed

### DIFF
--- a/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_admin_add_object_property.n3
+++ b/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_admin_add_object_property.n3
@@ -4,6 +4,7 @@
 @prefix access: <https://vivoweb.org/ontology/vitro-application/auth/vocabulary/> .
 
 :AdminAddObjectPropertyValueSet access:value
+    <http://vitro.mannlib.cornell.edu/ns/vitro/public#mainImage> ,
     <http://aims.fao.org/aos/geopolitical.owl#hasBorderWith> ,
     <http://aims.fao.org/aos/geopolitical.owl#hasMember> ,
     <http://aims.fao.org/aos/geopolitical.owl#isAdministeredBy> ,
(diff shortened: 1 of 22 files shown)


balmas (approved):
I can verify that this works. I tested for all role types (admin, curator, editor, self-editor) and confirm that before loading the new property files only the root user was able to add/edit/delete the profile image and after loading the new property files each role type had the ability to add/edit/delete the profileImage (where appropriate for that role/profile combination). Tested mainly in wilma theme but did a quick test with the self-editor role type in the other themes as well).

It would be ideal if there were an integration test to validate that all of the properties needed to support the default functionality of the webapp are there. It looks like the Selenium tests haven't been updated/used in years.  Not sure if there is another approach that would be easy to implement.

gneissone:
I built with the changes to VIVO and Vitro but still cannot see the image when logged out. @balmas were you also able to view the images when completely logged out and after doing a hard refresh on the page? Or in incognito/private mode?

balmas:
Oh, that's a good catch. I didn't try to do a hard refresh after logout. You are correct, they are not visible in that case.

balmas (changes requested):
- Changing Review status to Request changes, as noted by @gneissone  the uploaded images aren't visible to public users

litvinovg:
Thanks for catching that! I updated permission sets. Please test one more time.

balmas (approved):
Retested and confirm that can now view the profile images as a public (non-logged in user), and image editing capabilities remain correct for all other user role types.

gneissone (approved):
Tested and appears to work as expected now, thanks!

Issue #4069, opened 2025-04-16
Capability Map doesn't load due to MIME type error


gneissone:
**Describe the bug**
Note, I haven't been able to reproduce this except on a client's VIVO instance. It might be specific to certain server configurations, but the solution should be applicable to everyone (and best practice).

**To Reproduce**
1. Go to Capability Map
2. Search for known concept loaded in VIVO
3. Observe that Capability Map does not load
4. Open developer tools console and observe error: 
`Refused to execute script from 'https://vivo.mydomain.edu/vis/capabilitymap#Astronomy%20&%20Astrophysics%7C10%7C0' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.`

**Expected behavior**
Capability Map works

**Environment (please complete the following information):**
 - OS: Linux
 - Browser: Chrome
 - Tomcat version: 9
 - VIVO version: 1.14.1
 - Apache Solr: 8.x

**Additional context**
I will submit a small PR that I think fixes the issue

litvinovg:
I can't reproduce the issue too.
Could you check if there is any cross origin error on the instance where the problem was noticed?

brianjlowe:
Resolved by https://github.com/vivo-project/VIVO/pull/4070

PR #4070, opened 2025-04-16
VIVO-4069 - Add javascript MIME type to ipretResults responses

gneissone:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/4069)**: https://github.com/vivo-project/VIVO/issues/4069

Sets the MIME type for a couple specific responses from the Ajax Visualization Controller which are formatted as javascript instead of plaintext or JSON.

# What's new?
Functionality should remain the same, but you should be able to confirm the MIME type is set correctly by looking at the response in your developer console in the browser.
Screenshot 2025-04-16 at 16 54 43

# How should this be tested?
Observe the Content-Type when making a call to http://localhost:8080/vivo/visualizationAjax?vis=capabilitymap&query=Virology&callback=ipretFullResults (You will need to replace the query with a valid concept in your VIVO database)
The Content-Type before the change will be "text/html"
The Content-Type after the change will be "text/javascript"

Note that application/javascript is also a valid MIME type for javascript, but apparently text/javascript has broader compatibility.

# Additional Notes:

# Interested parties
@VIVO-project/vivo-committers

# Reviewers' expertise
**Please add any new expertise in the list which might be needed for reviewing your PR or remove any of the listed if it is not needed.**

1. Java
2. JavaScript

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/visualization/AjaxVisualizationController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/visualization/AjaxVisualizationController.java
@@ -13,6 +13,7 @@ import javax.servlet.http.HttpServletResponse;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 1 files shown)


brianjlowe (approved):
Confirmed that these are now served with the proper content type. Thanks for fixing this, and apologies for the long review delay.

Issue #4077, opened 2025-04-29
Docker instructions in the README need updating


balmas:
**Describe the bug**
The Docker build instructions in the README are no longer sufficient to standup a VIVO instance for testing.  I think this is likely due to changes introduced for #3034 which decouple the build from the installation.  The problem is that the documentation now misses a step about how to to setup your local VIVO home directory in order to run run the Docker image.

The instructions currently say:

> 
> Before building VIVO, you will also need to clone (and switch to the same branch, if other than main) of [Vitro](https://github.com/vivo-project/Vitro). The Vitro project must be cloned to a sibling directory next to VIVO so that it can be found during the build.
> 
> Build and start VIVO.
> 
> In VIVO (with Vitro cloned alongside it), run:
> mvn clean package -s installer/example-settings.xml
> docker-compose up

**To Reproduce**

In a clean environment with no local vivo home directory

```
git clone git@github.com:vivo-project/VIVO.git
git clone git@github.com:vivo-project/Vitro.git
cd VIVO
mvn clean package -s installer/example-settings.xml
docker-compose up --build 
```

Error occurs because this mvn target no longer produces a vivo.war file

```
=> ERROR [vivo 3/4] COPY ./installer/webapp/target/vivo.war /usr/local/t  0.0s
------
 > [vivo 3/4] COPY ./installer/webapp/target/vivo.war /usr/local/tomcat/webapps/ROOT.war:
------
failed to solve: failed to compute cache key: failed to calculate checksum of ref b5623dd3-d957-485b-a9af-31892ffb7038::7gyvmdysa8x8zu67fh2nwaa34: "/installer/webapp/target/vivo.war": not found
```

If instead I run the the Docker build command that is in the github docker workflow file, I do get a vivo war file, but then I get a different error on startup:

```
 mvn clean install -Dvivo-dir=/usr/local/vivo/home -Dapp-name=vivo -Ddefault-theme=wilma
docker-compose up --build 
```

Error:
```
me=3
vivo       | Copying home bin directory to /usr/local/vivo/home/bin
vivo       | cp: cannot stat '/vivo-home/bin': No such file or directory
vivo exited with code 1
```

**Expected behavior**
I would expect a valid instance of VIVO to be running 

**Environment (please complete the following information):**
 - OS: Mac 14.7.4 (23H420)
 - VIVO version main @ 032c424a4af6dc408ca6e68c6d9a62a271e44155
[...]

Issue #4047, opened 2025-02-02
tomcat image referenced in Dockerfile does not work on Mac Sonoma 14.5


balmas:
**Describe the bug**
The default docker image for VIVO does not work on MacOS Sonoma 14.5

**To Reproduce**
try to run the vivo:latest image, or build a new image following the instructions in the readme. Tomcat fails to start with a null pointer exception.  
To confirm that this wasn't a problem with the VIVO war file, I created an empty Docker file with just the tomcat image that runs catalina run . Same failure occurs.  Switching to another image (tomcat:9.0.48-jdk11-openjdk-slim) resolved the problem and 
I was able to run VIVO.

**Screenshots**

```
vivo       | 02-Feb-2025 21:14:27.829 INFO [main] org.apache.catalina.core.AprLifecycleListener.initializeSSL OpenSSL successfully initialized [OpenSSL 1.1.1n  15 Mar 2022]
vivo       | 02-Feb-2025 21:14:27.867 SEVERE [main] org.apache.catalina.startup.Catalina.load Error initializing Catalina
vivo       | 	org.apache.catalina.LifecycleException: Failed to initialize component [StandardServer[8005]]
vivo       | 		at org.apache.catalina.util.LifecycleBase.handleSubClassException(LifecycleBase.java:440)
vivo       | 		at org.apache.catalina.util.LifecycleBase.init(LifecycleBase.java:139)
vivo       | 		at org.apache.catalina.startup.Catalina.load(Catalina.java:724)
vivo       | 		at org.apache.catalina.startup.Catalina.load(Catalina.java:746)
vivo       | 		at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
vivo       | 		at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
vivo       | 		at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
vivo       | 		at java.base/java.lang.reflect.Method.invoke(Method.java:566)
vivo       | 		at org.apache.catalina.startup.Bootstrap.load(Bootstrap.java:305)
vivo       | 		at org.apache.catalina.startup.Bootstrap.main(Bootstrap.java:475)
vivo       | 	Caused by: java.lang.NullPointerException
[...]

PR #4078, opened 2025-05-05
Docker build fixes

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/4077)**

Fixes docker build

# Interested parties
@balmas @VIVO-project/vivo-committers

# Reviewers' expertise

1. Docker

# Reviewers' report template
## General comment
A reviewer should provide here comments and suggestions for requested changes if any.
## Testing
A reviewer should briefly describe here how it was tested
## Code reviewing
A reviewer should briefly describe here which part was code reviewed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,10 @@ ENV VIVO_HOME=${VIVO_HOME}
 ENV SOLR_URL=${SOLR_URL}
 
 RUN mkdir -p ${VIVO_HOME}
+RUN mkdir /vivo-home
 
 # Copy VIVO home onto image for backup, initialization, and reset
+COPY ./home/src/main/resources/config /vivo-home/config
 COPY ./installer/webapp/target/vivo.war /usr/local/tomcat/webapps/ROOT.war
 
 COPY start.sh /start.sh
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Build and start VIVO.
 
 1. In VIVO (with Vitro cloned alongside it), run:
 ```
-mvn clean package -s installer/example-settings.xml
+mvn clean package -s installer/docker-example-settings.xml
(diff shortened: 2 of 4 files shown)


balmas:
This almost fixes the problem but not 100%.  There is still a discrepancy between the installation instructions in the README . The README instructions you to run `mvn clean package -s installer/example-settings.xml` before running `docker-compose`.  As of right now, that will fail because the path for the resulting vivo.war as defined in that settings file is `/usr/local/tomcat/webapps`, while the Dockerfile expects the war to be in the local `./installer/webapp/target/vivo.war`.  So now that I know that, I am able to just get it working by copying the war file where it needs to be, but we really either need to document that step in the README, or better yet, provide an alternate example-settings.xml that is specifically meant for use with a Docker installation.  I'm not entirely sure of all the motivations behind the commits that introduced that change (https://github.com/vivo-project/VIVO/commit/7c2295f1dbee43d37668533244f4d6b85d035577 and https://github.com/vivo-project/Vitro/commit/53ca1d8558a2fc30c13a3fcd3a5a7bb29b21587f) so I don't feel confident that I can suggest the right solution here.

litvinovg:
@balmas I created installer settings for docker build and updated README. Could you try it once again?

balmas:
Yes this works perfectly. (I still have to adjust my Dockerfile to use a different base image to work on Mac Sonoma Apple Me Pro as reported in #4047 and adjust my docker-compose.yml to use a locally built Solr image because the VIVO solr image doesn't work on the Apple chip but those are separate issues. ). 

I think we can close this item because the problem reported in it is now fixed.

balmas (approved):

chenejac (approved):

Issue #4105, opened 2025-10-20
VIVO Harvester as part of VIVO


chenejac:
**Is your feature request related to a problem? Please describe.**
Data ingestion process wasn't offered as out-of-the-box feature of VIVO. 

**Describe the solution you'd like**
Deployment process via maven or docker should offer configuration to include VIVO-Harvester.
UI of VIVO/Vitro should offer page where input parameters for VIVO-Harvester might be defined: VIVO-Harvester project which define ETL process, configuration parameters are displayed in UI, and xslt file can be downloaded and uploaded, schedule of ETL process - one-time or occurrence frequency.

PR #4136, opened 2026-02-25
Feature/harvester integration

ivanmrsulja:
Issue [4105](https://github.com/vivo-project/VIVO/issues/4105)

Linked [Vitro PR](https://github.com/vivo-project/Vitro/pull/509)
Linked [VIVO-Harvester PR](https://github.com/vivo-project/VIVO-Harvester/pull/69)

Implemented VIVO - VIVOHarvester integration using configuration file and dynamic form rendering.

# What's new?
- edu.cornell.mannlib.vivo.harvest
- webapp/src/main/webapp/templates/freemarker/body/harvester/harvest-dashboard.ftl
- webapp/src/main/webapp/templates/freemarker/body/harvester/task-details.ftl
- webapp/src/main/webapp/css/harvester/dashboard.css

# How should this be tested?
In `runtime.properties`, the following should be set:
- `harvester.directory`=directory where VIVOHarvester is installed
- `harvester.configuration`=example-scripts/bash-scripts/full-harvest-examples/examples/export-config/export-config.json
- `workflow.scheduler.scan`=0 * * * * * - for testing only, once per day in production environments
- `workflow.log.directory`=/tmp

Additionally, VIVOHarvester should be built from [this PR](https://github.com/vivo-project/VIVO-Harvester/pull/69) and Vitro should be built from [this PR](https://github.com/vivo-project/Vitro/pull/509).

After that, simply logging in as `admin` and navigating to `/etlWorkflows` should present all relevant functionality to you (task running, scheduling, scheduled task CRUD and current/scheduled task log CRUD).

# Interested parties
@chenejac 

# Reviewers' expertise

1. Java
1. HTML, CSS, JavaScript
1. FreeMarker
1. Natural language knowledge
    1. English
    2. German
    3. Spanish
    4. French
    5. Portuguese
    6. Russian
    7. Serbian

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vivo/harvest/HarvestContextSetup.java
+++ b/api/src/main/java/edu/cornell/mannlib/vivo/harvest/HarvestContextSetup.java
@@ -1,9 +1,12 @@
 package edu.cornell.mannlib.vivo.harvest;
 
+import static edu.cornell.mannlib.vitro.webapp.modelaccess.ModelAccess.WhichService.CONTENT;
+
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
 
 import javax.servlet.ServletContext;
 import javax.servlet.ServletContextEvent;
@@ -11,6 +14,9 @@ import javax.servlet.ServletContextListener;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import edu.cornell.mannlib.vitro.webapp.config.ConfigurationProperties;
+import edu.cornell.mannlib.vitro.webapp.dao.jena.BlankNodeFilteringModelMaker;
+import edu.cornell.mannlib.vitro.webapp.modelaccess.ModelAccess;
+import edu.cornell.mannlib.vitro.webapp.rdfservice.RDFService;
 import edu.cornell.mannlib.vivo.harvest.configmodel.ExportConfig;
 import edu.cornell.mannlib.vivo.harvest.contextmodel.HarvestContext;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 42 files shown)


api/src/main/java/edu/cornell/mannlib/vivo/harvest/HarvestContextSetup.java
```diff
@@ -0,0 +1,115 @@
...
+
+        SchedulerManager.scheduleTasks(
+            new ScheduledHarvestExecutor()
+        );
+    }
+
+    @Override
+    public void contextDestroyed(ServletContextEvent servletContextEvent) {
+        HarvestContext.modules.clear();
+        HarvestJobExecutor.shutdown();
```
brianjlowe, line 113: It looks like SchedulerManager.shutdown() also needs to be called here. Without that, Tomcat is never shuts down.

api/src/main/java/edu/cornell/mannlib/vivo/scheduler/SchedulerManager.java
```diff
@@ -0,0 +1,111 @@
...
+            long delay = nextRun.getTime() - now.getTime();
+            scheduler.schedule(() -> {
+                task.run();
+                scheduleNextRun(task, cron);
+            }, delay, TimeUnit.MILLISECONDS);
+        }
+    }
+
+    public static void shutdown() {
+        scheduler.shutdown();
```
brianjlowe, line 109: Consider scheduler.shutdownNow() here? I found that the regular shutdown() still delayed my shutdown of Tomcat by several minutes.

brianjlowe (changes requested):
Thank you, Ivan, for this substantial body of work. The harvest module GUI is very nice! I made a few inline comments, and have just a few others here:

1. The dropdown for selecting the named graph for the harvest is excellent, but it looks like the list only updates when Tomcat is restarted. I think we'll want it to show the current list so as to respond to new named graphs that users create in the ingest tools menu. (For example, I went to create a new named graph for OpenAlex and wanted to use it in an OpenAlex harvest, but it wasn't offered to me until I populated it and restarted Tomcat. That won't be possible in production environments.)

2. There's a little mix of languages on the scheduled tasks tab:
Image

3. Running and stopping workflows works well, but the process breaks down at the final SPARQL update, so I don't actually have data in my graphs. How are the email and password configured for the update API? Here are the last lines from an OpenAlex harvest:
```2026-09-03 15:56:28.978 INFO  [o.v.h.s.SparqlUpdate] SparqlUpdate: Start
HTTP/1.1 403 
<H1>403 email/password combination is not valid</H1>
2026-09-03 15:56:30.037 INFO  [o.v.h.s.SparqlUpdate] SparqlUpdate: End
Import Statistics:
```

Interestingly, if I have a harvest that generates more than a small amount of data, I get a broken pipe instead of the 403. I'm guessing that might be because it's trying to insert data in one chunk that exceeds my Tomcat's maxPostSize, so addressing that may be out of scope for this PR.

4. I scheduled an OpenAlex harvest that had a start date of 2018-01-01 and and end date of 2018-01-31, but the scheduled task no longer shows me the start and end dates for the harvest. Will it continue to use the two original dates I entered?

PR #4125, opened 2025-11-28
Feature/orcid push

ivanmrsulja:
Issues: #4074, #4075

linked Vitro PR: [504](https://github.com/vivo-project/Vitro/pull/504)

# What does this pull request do?
Integrates ORCID push workflow using Member API credentials.

# What's new?
- edu.cornell.mannlib.vivo.orcid.controller.OrcidDisallowPushHandler
- edu.cornell.mannlib.vivo.orcid.controller.OrcidAllowPushHandler
- edu.cornell.mannlib.vivo.orcid.controller.OrcidMemberCallbackHandler
- edu.cornell.mannlib.vivo.orcid.export
- edu.cornell.mannlib.vivo.orcid.service
- edu.cornell.mannlib.vivo.orcid.util

# How should this be tested?
Configure VIVO ORCID credentials in `runtime.properties` file (Ask @chenejac for sandbox credentials). Navigate to researcher profile when logged in as researcher (you can bind researcher account to any person entity). Click on `Allow ORCID push` and complete the OAuth2 workflow (keep in mind that you will have to create ORCID Sandbox account beforehand, your regular ORCID account won't work). After this, periodic dissemination of research information should be performed to your ORCID profile.

# Interested parties
@chenejac 

# Reviewers' expertise
**Please add any new expertise in the list which might be needed for reviewing your PR or remove any of the listed if it is not needed.**

Candidates for reviewing this PR should have some of the following expertises:
1. Java
1. HTML, CSS, JavaScript
1. FreeMarker
1. SPARQL

PR #4145, opened 2026-04-09
ORCiD authentication

### LINKED ISSUE #4071: ORCID authentication (login) within VIVO
Enabling VIVO users to register or sign into VIVO system using their ORCID credentials can save them time and effort. Moreover, VIVO users don't have to keep track of multiple usernames and passwords, and it is possible to immediately obtain and align an authenticated ORCID iD to a person/researcher profile. At the same time, it is possible to request permission to read or update VIVO users' ORCID record and linked publications.

- Aligned with the best practice described here - https://info.orcid.org/documentation/integration-guide/sign-in-using-orcid-credentials/

**Additional context**
It has been already implemented in [OpenVIVO](https://github.com/OpenVIVO/VIVO)

bkampe:
**Sign-in using ORCID credentials.**
* * *

**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/4071)**: (ORCID authentication (login) within VIVO #4071)

This pull request adds the posibility to sign-in using ORCID credentials.

# What's new?
Added java class OrcidAuthController.java which contains main functionality to log in with ORCID credentials. With a click on "Login via ORCID" the user gets redirected to the ORCID login page. After entering the crendentials ORCID redirects to a predefined callback-URL.

# How should this be tested?
To test this feature you need to
* Set up an publicly available VIVO instance including this pull request
* Register a API client at the ORCID page to get client credentials for your VIVO instance
* Register your redirect URLs at ORCID developer tools
* Add the clients credentials to runtime.properties. e.g.:
`auth.orcid.clientId             = APP-0FHD0100VVK2E4FP`
`auth.orcid.clientPassword       = bed12525-xxxx-xxxx-xxxx-xxxxxxxxxxxxx`
`auth.orcid.callbackUrl          = http://kampe.ddns.net:62808/orcid_auth_war/orcidAuth/callback`

# Additional Notes:
For registering an API client and setting callback URLs see https://info.orcid.org/documentation/api-tutorials/api-tutorial-get-and-authenticated-orcid-id/

# Interested parties
@chenejac @hauschke


api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1135 @@
...
+        userAccount.setExternalAuthId(orcidToken.orcid);
+        userAccount.setPasswordChangeRequired(false);
+        userAccount.setPasswordLinkExpires(0);
+        userAccount.setExternalAuthOnly(true);
+        userAccount.setLoginCount(0);
+        userAccount.setStatus(UserAccount.Status.ACTIVE);
+        userAccount.setPermissionSetUris(Collections.singleton(PermissionSets.URI_SELF_EDITOR));
+
+        UserAccountsDao userAccountsDao = vreq.getWebappDaoFactory().getUserAccountsDao();
+        userAccountsDao.insertUserAccount(userAccount);
```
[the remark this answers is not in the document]
bkampe: It looks to me like the cause isn't in OrcidAuthController but rather in [UserAccountsDaoJena.java](https://github.com/vivo-project/Vitro/blob/a2367bde353092f72f97aa2c3c0892cf4cecd7d5/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/UserAccountsDaoJena.java#L474), [NewURIMakerVitro.java](https://github.com/vivo-project/Vitro/blob/a2367bde353092f72f97aa2c3c0892cf4cecd7d5/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/NewURIMakerVitro.java#L21), and [UserAccount.java](https://github.com/vivo-project/Vitro/blob/a2367bde353092f72f97aa2c3c0892cf4cecd7d5/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/UserAccount.java#L139), where the insecure java.util.Random is used instead of java.security.SecureRandom.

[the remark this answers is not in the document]
bkampe: Maybe https://github.com/vivo-project/Vitro/pull/519 solves the problem.

api/pom.xml
```diff
@@ -45,7 +45,11 @@
             <artifactId>orcid-api-client</artifactId>
             <version>0.6.4</version>
         </dependency>
-
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
```
chenejac: It would be better to use jackson instead of introducing the new dependency. Hopefully, it might decrease OrcidAuthController boilerplate code at the bottom.

chenejac (changes requested):
@bkampe thanks for this. You might be right about security warning, the issue might be in the Vitro codebase. However, we will know that once we merge your Vitro pull request on the main branch. Meanwhile, let's try to clean up code here. I left one comment about gson dependency. Please try to address that.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
+/* $This file is distributed under the terms of the license in /doc/license.txt$ */
+
+package org.vivoweb.webapp.controller.freemarker;
+
+// 1. STANDARD_JAVA_PACKAGE Group (Top priority)
```
brianjlowe: These comments can be removed.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
...
+    private static final String API_VERSION = "v3.0";
+    private static final String CONFIGURATION_ORCID_API = "orcid.api";
+    private static final String CONFIGURATION_AUTH_ORCID_CALLBACK_URL = "auth.orcid.callbackUrl";
+    private static final String CONFIGURATION_AUTH_ORCID_CLIENT_PASSWORD = "auth.orcid.clientPassword";
+    private static final String CONFIGURATION_AUTH_ORCID_CLIENT_ID = "auth.orcid.clientId";
+    private static final String BIBO_DOI = "http://purl.org/ontology/bibo/doi";
+    private static final String CODE_PARAM = "code";
+    private static final String CALLBACK = "callback";
+    private static final String CLAIM_DOIS_FTL = "claimDOIs.ftl";
+    private static final String PROFILE_ERROR_FTL = "profileError.ftl";
```
brianjlowe: This, along with VIVO_REJECTED_DOI and many of the methods in the class show in the IDE as unused. This looks to be due to new profile creation and DOI claiming being out of scope for this feature. Please remove the unused code.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
...
+        apiVersion = API_VERSION;
+
+        if (clientSecret != null && callbackUrl != null && clientId != null && apiType != null) {
+            if (apiType.equals(SANDBOX)) {
+                apiPrefix = "sandbox.";
+            } else {
+                apiPrefix = "";
+            }
+        } else {
+            String errstr = "ORCiD variables missing in properties file!";
```
brianjlowe: ORCID, when referring generally to the organization or service, should be written in all capital letters. The ID should be referred to as an "ORCID iD", with the lowercase "i" occurring only in the second word, or as an "ORCID identifier". https://info.orcid.org/brand-guidelines/ .
It also might be helpful to mention runtime.properties specifically in this message and to list the names of the properties that are missing.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
...
+            String location = getRedirectLocation();
+            return new DirectRedirectResponseValues(location);
+        }
+        try {
+            verifyOrcidContextIsAvailable();
+            verifyUserIsNotLoggedIn(vreq);
+            OrcidTokenResponse orcidToken = getOrcidToken(vreq);
+            if (userAccountExists(vreq, orcidToken)) {
+                return login(vreq, orcidToken);
+//          } else {
```
brianjlowe: Remove commented code.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
...
+    }
+
+    private ResponseValues login(VitroRequest vreq, OrcidTokenResponse orcidToken) {
+        try {
+            OrcidPerson orcidBio = getPersonDetails(orcidToken);
+            String familyName = getFamilyName(orcidBio);
+            String givenName = getGivenName(orcidBio);
+            UserAccount userAccount = getAuthenticator(vreq).getAccountForExternalAuth(orcidToken.orcid);
+            String profileUri = getProfileUri(vreq, userAccount);
+//          if (profileUri == null) {
```
brianjlowe: Remove commented code.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
...
+//              profileUri = createProfile(vreq, orcidToken, orcidBio, familyName, givenName, userAccount);
+//          }
+
+            getAuthenticator(vreq).recordLoginAgainstUserAccount(userAccount,
+                    LoginStatusBean.AuthenticationSource.EXTERNAL);
+            String[] dois = getOrcidDOIs(vreq.getRDFService(), profileUri, orcidToken);
+
+//          if (shouldGoToClaimDoi(dois)) {
```
brianjlowe: Remove

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
...
+            if (profileUri == null) {
+                return goToHomePage();
+            } else {
+                return goToUserProfile(vreq, profileUri);
+            }
+
+
+        } catch (Authenticator.LoginNotPermitted e) {
+            return new TemplateResponseValues(NOT_AUTHENTICATED_FTL);
+//      } catch (NotEnoughInfoForNewProfileException e) {
```
brianjlowe: Remove

webapp/src/main/webapp/themes/nemo/i18n/vivo_all_en_US.properties
```diff
@@ -33,6 +33,7 @@ menu_welcomestart = Welcome
 
 menu_logout = Log out
 menu_login = Log in
+menu_login_orcid = Log in via ORCiD
```
brianjlowe: ORCiD -> ORCID

webapp/src/main/webapp/themes/wilma/templates/identity.ftl
```diff
@@ -46,6 +46,7 @@
 
             <#else>
                 <li role="listitem"><a class="log-out" title="${i18n().menu_loginfull}" href="${urls.login}">${i18n().menu_login}</a></li>
+                <li role="listitem"><a class="log-out" title="${i18n().menu_loginfull}" href="http://localhost:8080/orcid_auth_war/orcidAuth">${i18n().menu_login_orcid}</a></li>
```
brianjlowe: Hardcoded localhost path here

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,1170 @@
...
+    private String apiVersion;
+    private ObjectMapper mapper;
+
+    @Override
+    public void init() throws ServletException {
+        super.init();
+        mapper = new ObjectMapper();
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        mapper.setVisibility(PropertyAccessor.FIELD, Visibility.ANY);
+        ConfigurationProperties configProperties = ConfigurationProperties.getBean(getServletContext());
```
brianjlowe: Deprecated method here. Consider ConfigurationProperties.getInstance();

home/src/main/resources/rdf/i18n/en_US/interface-i18n/firsttime/vivo_UiLabel_wilma.ttl
```diff
@@ -832,6 +832,15 @@ uil-data:menu_login.VIVO.wilma
         uil:hasPackage  "VIVO-languages" ;
         uil:hasTheme    "wilma" .
 
+uil-data:menu_login_orcid.VIVO.wilma
+        rdf:type         owl:NamedIndividual ;
+        rdf:type         uil:UILabel ;
+        rdfs:label       "Log in via ORCiD"@en-US ;
```
brianjlowe: ORCiD -> ORCID

brianjlowe (changes requested):
Thank you for this PR. I've noted some lines where minor changes should be made. The main thing to do at this stage is to remove any unused code, template files, etc. related to the DOI claiming and profile creation functionality that is not used. This should greatly simplify the PR and any remaining review.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
bkampe:
```suggestion
            String errstr = "ORCID variables missing in runtime.properties file!" +
                    "\nPlease check if auth.orcid.clientId, auth.orcid.clientPassword and auth.orcid.callbackUrl are set";
```

webapp/src/main/webapp/themes/wilma/templates/identity.ftl
bkampe: To build the url in the standard way (urls.orcidAuth) I added the ORCID path to the URLBuilder in vitro:
https://github.com/vivo-project/Vitro/pull/523

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,501 @@
...
+    private void verifyOrcidContextIsAvailable() throws OrcidNotConfiguredException {
+        OrcidClientContext orcidClientContext = OrcidClientContext.getInstance();
+        if (orcidClientContext == null) {
+            throw new OrcidNotConfiguredException();
+        }
+    }
+
+    private ResponseValues login(VitroRequest vreq, OrcidTokenResponse orcidToken) {
+        try {
+            OrcidPerson orcidBio = getPersonDetails(orcidToken);
```
brianjlowe: The orcidBio variable appears not to be used. When this line is taken out, getPersonDetails() and additional classes also become unused.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,501 @@
...
+            return new TemplateResponseValues(ORCID_NOT_CONFIGURED_FTL);
+        } catch (UnexpectedLoggedInUserException e) {
+            if (e.getUrl() == null) {
+                return new TemplateResponseValues(UNKNOWN_PROFILE_FTL);
+            } else {
+                return new DirectRedirectResponseValues(e.getUrl());
+            }
+        } catch (AuthenticationException e) {
+            String message = e.getMessage();
+            Map<String, Object> templateValues = new HashMap();
```
brianjlowe:
```suggestion
            Map<String, Object> templateValues = new HashMap<String, Object>();
```

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,501 @@
...
+    private static final String ORCID_NOT_CONFIGURED_FTL = "orcidNotConfigured.ftl";
+
+    private static final String MESSAGE = "message";
+
+    private Log log = LogFactory.getLog(this.getClass());
+    private String clientId;
+    private String clientSecret;
+    private String callbackUrl;
+    private String apiPrefix;
+    private String apiVersion;
```
brianjlowe: The value of apiVersion is not read. Was it intended to be checked?

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,501 @@
...
+    private String getRedirectLocation() {
+        return "https://" + apiPrefix + "orcid.org/oauth/authorize" + "?client_id=" + clientId
+                + "&response_type=code" + "&scope=/authenticate" + "&redirect_uri=" + callbackUrl;
+    }
+
+    private Authenticator getAuthenticator(HttpServletRequest req) {
+        return Authenticator.getInstance(req);
+    }
+
+    private OrcidPerson getPersonDetails(OrcidTokenResponse token) {
```
brianjlowe: No longer used if line 165 (unused orcidBio) is removed above.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    private static class OrcidRecord {
+        @JsonProperty("orcid-identifier")
+        OrcidIdentifier orcidIdentifier;
+
+        @JsonProperty("person")
+        OrcidPerson person;
+    }
+
```
brianjlowe: This and several of the related classes below become unused once getPersonDetails() is removed.

brianjlowe (changes requested):
Many thanks for the updated PR(s). Looks like it's almost ready. It appears that just a few more methods and inner classes can now be removed since the orcidBio value is not being used; see inline comments.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
bkampe: It was used in getPersonDetails(). Not needed anymore.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,348 @@
...
+            return new DirectRedirectResponseValues(location);
+        }
+        try {
+            verifyOrcidContextIsAvailable();
+            verifyUserIsNotLoggedIn(vreq);
+            OrcidTokenResponse orcidToken = getOrcidToken(vreq);
+            if (userAccountExists(vreq, orcidToken)) {
+                return login(vreq, orcidToken);
+            } else {
+                return null;
```
brianjlowe: If you log in with an ORCID iD that doesn't correspond to an existing account, you get "There was an error in the system" because of the null returned here. This needs to send the user somewhere; for now, it may be fine to use the same UNKNOWN_PROFILE_FTL template used on line 131. While this page is too simplistic, it wasn't introduced in this PR so it could be improved as a separate issue. The current stack trace logged is 
```2026-06-02 17:16:01,224 ERROR [FreemarkerHttpServlet] java.lang.NullPointerException: Cannot invoke "edu.cornell.mannlib.vitro.webapp.controller.freemarker.responsevalues.ResponseValues.getStatusCode()" because "values" is null
java.lang.NullPointerException: Cannot invoke "edu.cornell.mannlib.vitro.webapp.controller.freemarker.responsevalues.ResponseValues.getStatusCode()" because "values" is null
        at edu.cornell.mannlib.vitro.webapp.controller.freemarker.FreemarkerHttpServlet.doResponse(FreemarkerHttpServlet.java:244)
        at edu.cornell.mannlib.vitro.webapp.controller.freemarker.FreemarkerHttpServlet.doGet(FreemarkerHttpServlet.java:111)```

File: webapp/src/main/webapp/templates/freemarker/body/orcidAuth/claimDOIs.ftl [hunk 20]
```diff
@@ -0,0 +1,35 @@
+<style>
```
brianjlowe: It looks like this template can be removed?

webapp/src/main/webapp/templates/freemarker/body/orcidAuth/noProfile.ftl
```diff
@@ -0,0 +1,4 @@
+<h2>ORCiD Profile is not Public</h2>
```
brianjlowe: It looks like this template can be removed?

webapp/src/main/webapp/templates/freemarker/body/orcidAuth/usageAgreement.ftl
```diff
@@ -0,0 +1,10 @@
+    <p>You have to accept this agreement to register an account</p>
```
brianjlowe: It looks like this template can be removed?

webapp/src/main/webapp/themes/wilma/templates/identity.ftl
```diff
@@ -46,6 +46,7 @@
 
             <#else>
                 <li role="listitem"><a class="log-out" title="${i18n().menu_loginfull}" href="${urls.login}">${i18n().menu_login}</a></li>
+                <li role="listitem"><a class="log-out" title="${i18n().menu_loginfull}" href="${urls.orcidAuth}">${i18n().menu_login_orcid}</a></li>
```
brianjlowe: The link needs to be included in the interface only when the necessary ORCID properties are configured in runtime.properties. Currently, when they are not, the link is still presented which leads to the error below being presented to the user. Additionally, I don't think identity.ftl is the place to put the link. Other external authentication links are added to the login widget; the ORCID authentication option should be presented there when ORCID is configured.
```
HTTP Status 500 – Internal Server Error

Type Exception Report

Message ORCID variables missing in runtime.properties file!

Description The server encountered an unexpected condition that prevented it from fulfilling the request.

Exception

javax.servlet.ServletException: ORCID variables missing in runtime.properties file!
```

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,348 @@
...
+/**
+ *
+ */
+@WebServlet("/orcidAuth/*")
+public class OrcidAuthController extends FreemarkerHttpServlet {
+
+    private static final String JSON_TOKEN_PARAM = "json_token";
+    private static final String SANDBOX = "sandbox";
+    private static final String CONFIGURATION_ORCID_API = "orcid.api";
+    private static final String CONFIGURATION_AUTH_ORCID_CALLBACK_URL = "auth.orcid.callbackUrl";
```
brianjlowe: Can this be updated to use the existing orcid.clientId, orcid.clientSecret and orcid.webappBaseUrl properties instead of introducing new ones? If it is necessary to have a second set of properties for the authentication, they need to be added to example.runtime.properties with the appropriate explanatory text.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,348 @@
...
+        } catch (AuthenticationException e) {
+            String message = e.getMessage();
+            Map<String, Object> templateValues = new HashMap<String, Object>();
+            templateValues.put(MESSAGE, message);
+            return new TemplateResponseValues(NOT_AUTHENTICATED_FTL, templateValues);
+        }
+    }
+
+    private boolean userAccountExists(VitroRequest vreq, OrcidTokenResponse orcidToken) {
+        return getAuthenticator(vreq).getAccountForExternalAuth(orcidToken.orcid) != null;
```
brianjlowe: Perhaps this is how the feature was specified, but this puts undue restriction on the way user accounts are linked to foaf:Persons. It is quite likely that a user account will already be linked to a foaf:Person profile by some other arbitrary networkId value. It should be possible to enable ORCID authentication by adding the vivo:orcidId to the foaf:Person without changing the networkId to match. I think the logic here should be to retrieve a profile whose vivo:orcidId matches the orcidToken.orcid and then see if there is a linked user account, no matter what value is used to link that user account to the profile.

brianjlowe (changes requested):
Thanks for the revisions. I've now tested this on a public server and confirmed that the authentication does work when the ORCID iD is used directly in the networkId matching property, but please see my inline comment related to this. I've also noted a few additional things things that need some attention.

webapp/src/main/webapp/themes/wilma/templates/identity.ftl
bkampe: This is a good point, putting it into the login widget is way better. I put it in identity.ftl just as a quick and dirty proof of concept, but then it stayed there. Will take a look at the login widget.

bkampe: I added the functionality to add the ORCID login button to the login widget to my Vitro pull request:
https://github.com/vivo-project/Vitro/pull/523/changes/2852ee50ab6e6a9351562f354478ef1bd8356824

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
bkampe: Not sure why I did't use the existing ones, maybe a kind of "copy-paste-blindness"

bkampe: related Vitro commit
https://github.com/vivo-project/Vitro/pull/523/changes/1f322b6f175bd3b93ac1d156d367c7b779446930

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
bkampe: That's a good suggestion, but for me it needed some investigation. I finally added some functionality that should fulfill your idea. Please have a look.
https://github.com/vivo-project/VIVO/pull/4145/changes/cbac0c366999ad3ea0ceab436d89f10df1d1193a

brianjlowe (changes requested):
Thanks for these updates. I've re-tested the various scenarios and the functionality is working as expected. I think we're down to just two little improvements that will make this more robust with dirtier data, and then we should be ready to merge:

1. In getPersonUriByOrcidUri() the first person URI with a matching vivo:orcidId property is returned. This may lead to failed logins if there are duplicate persons in the triple store, because the first person returned may not be the one with a matching property associated with a user account. I would suggest incorporating the matching property from the self editing configuration into the query in getPersonUriByOrcidUri() so that only persons associated with user accounts are returned. If there are multiple of these, then taking the first seems fine. I would, however, also suggest adding "ORDER BY ?person" to that query so that the first one returned is always the same given the same data.
2. On line 160, https protocol is assumed for the ORCID URI. This is a good assumption for any recent data, but I think it would be nice to allow for the possibility that older http:// ORCID URIs might exist in some VIVOs and thus check for both protocols.

Otherwise, looking good. Thanks again.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,476 @@
...
+                }
+            }
+        }
+        UserAccount fallbackAccount = getAuthenticator(vreq).getAccountForExternalAuth(orcidToken.orcid);
+
+        return fallbackAccount;
+    }
+
+    private String getPersonUriByOrcidId(VitroRequest vreq, String orcidId) {
+        final List<String> personUris = new ArrayList<String>();
```
brianjlowe: This could be String personUri instead of an ArrayList, since you're only retrieving a limit of one value.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,476 @@
...
+                    + "  {\n"
+                    + "    ?person <" + ORCID_ID_PROPERTY_URI + "> <" + orcidHttps + "> .\n"
+                    + "  } UNION {\n"
+                    + "    ?person <" + ORCID_ID_PROPERTY_URI + "> <" + orcidHttp + "> .\n"
+                    + "  }\n"
+                    + "  ?person <" + matchingPropertyUri + "> ?matchingValue .\n"
+                    + "}\n"
+                    + "ORDER BY ?person\n"
+                    + "LIMIT 1";
+        } else {
```
brianjlowe: The else branch looks unnecessary, since authentication can't continue if there is no matching ID.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,476 @@
...
+            log.error("Error finding person by ORCID ID: " + orcidId, e);
+        }
+
+        if (personUris.isEmpty()) {
+            log.debug("getPersonUriByOrcidId: Query returned no results.");
+            return null;
+        }
+
+        return personUris.get(0);
+    }
```
brianjlowe: Missing blank line after this method.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,476 @@
...
+        return getLinkedUserAccount(vreq, orcidToken) != null;
+    }
+
+    private UserAccount getLinkedUserAccount(VitroRequest vreq, OrcidTokenResponse orcidToken) {
+        if (orcidToken == null || StringUtils.isEmpty(orcidToken.orcid)) {
+            log.debug("getLinkedUserAccount: orcidToken is null or orcid is empty");
+            return null;
+        }
+
+        String personUri = getPersonUriByOrcidId(vreq, orcidToken.orcid);
```
brianjlowe: Now that ?matchingValue is included in the first SPARQL query, I would suggest renaming/updating getPersonUriByOrcidId() to a return an object with both the person URI and the matching ID string. Then you can delete lines 164-170 (no need to retrieve the full individual) and you should be able to delete the getExternalAuthIdForIndividual() method as well, since you'll already have everything you need.

brianjlowe (changes requested):
The logic looks good to me; there are just a few opportunities now to simplify the code a bit, and then this should be good to go.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
bkampe: But it's used in the inner class `final protected void processQuerySolution(QuerySolution qs)`. I think a simple string variable could not be accessed from there? But it could be simplified to a one element string array ` String[] personUri = new String[1];`

brianjlowe: Thanks, you're right; I overlooked that. In that case, I would just leave it as is. I deleted my related comment below.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
brianjlowe: Taking another look at this, perhaps getPersonUriByOrcidId() could just become getExternalAuthIdForOrcidId() and return only the ?matchingValue instead of the ?person.

bkampe: Good suggestion. I was just modifying it to return both the person URI and the matching ID and then realized that the person ID is not needed in the calling function.

brianjlowe (approved):
Looks great. Many thanks for all your work on this.

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,425 @@
...
+
+    @Override
+    public void init() throws ServletException {
+        super.init();
+        mapper = new ObjectMapper();
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        mapper.setVisibility(PropertyAccessor.FIELD, Visibility.ANY);
+        ConfigurationProperties configProperties = ConfigurationProperties.getInstance();
+        clientId = configProperties.getProperty(CONFIGURATION_AUTH_ORCID_CLIENT_ID);
+        clientSecret = configProperties.getProperty(CONFIGURATION_AUTH_ORCID_CLIENT_PASSWORD);
```
chenejac, line 89: Can we use here OrcidClientContext properties? - OrcidClientContext.Setting.CLIENT_ID, OrcidClientContext.Setting.CLIENT_SECRET

api/src/main/java/org/vivoweb/webapp/controller/freemarker/OrcidAuthController.java
```diff
@@ -0,0 +1,425 @@
...
+        callbackUrl = configProperties.getProperty(CONFIGURATION_AUTH_ORCID_CALLBACK_URL) + "orcidAuth/callback";
```
chenejac, line 90: I think in [ORCID push data PR](https://github.com/vivo-project/VIVO/pull/4125) the hardcoded callback specified in OrcidClientContext (OrcidClientContext.Setting.CALLBACK_PATH) has been tried to be reused. It is allowed to have more than one callbackUrl/redirectUrl in [ORCID registration](https://info.orcid.org/documentation/api-tutorials/api-tutorial-get-and-authenticated-orcid-id/), but not sure what might be a good practice in our case taking into account maintainance and configuration? @bkampe Can you please investigate a little bit and write here in the comment your arguments? Here is an LLM response about advantages of single and multi callback URLs:

> Advantages of Using a Single Callback URL
> 
> Simpler ORCID Configuration
> 
> Only a single Redirect URI needs to be registered in the ORCID Developer Tools.
> Reduces the risk of configuration errors during migrations, deployments, or domain changes.
> 
> Easier Maintenance
> 
> All OAuth responses are handled by a single component or endpoint.
> Provides a centralized location for logging, auditing, monitoring, and debugging.
> 
> Alignment with Standard OAuth Practices
> 
> OAuth is designed to support multiple authorization flows through the use of the state parameter.
> Different operations (e.g., login, ORCID verification, data synchronization) can be distinguished and routed appropriately based on the value of the state parameter, while sharing the same callback URL.    

> Advantages of Using Multiple Callback URLs
> 
> Clear Separation of Concerns
> 
> Each ORCID-related flow (login, verification, data push) has its own dedicated endpoint.
> Business logic is naturally isolated per use case, which can make the code easier to understand in larger systems.
> 
> Simplified Request Handling
> 
> No need to inspect or route logic based on the state parameter for basic flow separation.
> Each callback directly maps to a single, well-defined purpose.
> 
> Improved Security Isolation (in some architectures)
> 
> Different callback endpoints can be secured independently.
> Useful when different flows require different validation rules, scopes, or permission checks.
> 
> Easier Integration with Modular or Microservice Architectures
> 
> Each service or module can expose its own callback endpoint.
> Reduces coupling between authentication flows and downstream processing logic.

chenejac (changes requested):
@bkampe please check my two comments. 

Also, I think we might discuss UI design of the feature. At the moment, when orcid.authEnabled is false, the login widget is the following one

Image

When orcid.authEnabled is true, the base login widget is the following one

Image

And it can be extended 

Image

The another option might be just to add `Log in via ORCID` button next to the green `Log in` button (right of that) at the first picture. 

At the moment, this is linked with Vitro code, but I think we should also discuss why Log in via ORCID UI is there, and expecially taking into account it uses labels defined in VIVO code base.

chenejac (approved):

chenejac (approved):

PR #4159, opened 2026-05-29
Add ror id

### LINKED ISSUE #4158: Support linking ROR identifier with an Organisation
ROR identifier is quite popular Organisation Unit PID. It is already present in vivo-ontology repository. Moreover, this is a requirement for adding affiliation/education into ORCID - [link](https://info.orcid.org/documentation/integration-guide/working-with-organization-identifiers/). 

Just add ROR identifier into vivo.owl file, and configure to be in Identity tab. 

**Describe alternatives you've considered**
Later we might add some verification whether ROR is in proper format (for instance based on regular expression). Also, there might be a code which is after adding ROR identfier enriching Organisation record in VIVO graph by using [ROR API](https://api.ror.org/v2/organizations/005gbx671) (adding country, city, other linked identifiers, etc.).

chenejac:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**: #4158 

Add support for cataloguing ROR identifier linked with an organisation.  

# What's new?
rorId data property copied from [vivo-ontology repository](https://github.com/vivo-ontologies/vivo-ontology/pull/56) into vivo.owl file used in the VIVO platform. 

The property is displayed in the Identity tab. The label for the property is ROR ID for all languages!

# How should this be tested?
1. Run VIVO with sample data
2. Log in as an admin
3. Open landing page of Harvard university in VIVO
4. Add ROR identifier - 03vek6s52
5. Check whether it is displayed

# Interested parties
Tag (@ mention) interested parties or, if unsure, @VIVO-project/vivo-committers

# Reviewers' expertise

1. Ontologies

### DIFF
--- a/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_public_publish_data_property.n3
+++ b/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_public_publish_data_property.n3
@@ -182,7 +182,7 @@
     vivo:rank ,
     vivo:reportId ,
     vivo:researchOverview ,
-    vivo:rodId ,
+    vivo:rorId ,
     vivo:seatingCapacity ,
     vivo:sponsorAwardId ,
     vivo:supplementalInformation ,
--- a/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_public_publish_data_property.n3
+++ b/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_public_publish_data_property.n3
@@ -182,6 +182,7 @@
     vivo:rank ,
     vivo:reportId ,
     vivo:researchOverview ,
+    vivo:rodId ,
     vivo:seatingCapacity ,
     vivo:sponsorAwardId ,
     vivo:supplementalInformation ,
--- a/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_admin_add_data_property.n3
+++ b/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_admin_add_data_property.n3
@@ -158,6 +158,7 @@
     <http://vivoweb.org/ontology/core#reportId> ,
     <http://vivoweb.org/ontology/core#researchOverview> ,
     <http://vivoweb.org/ontology/core#researcherId> ,
+    <http://vivoweb.org/ontology/core#rorId> ,
     <http://vivoweb.org/ontology/core#scopusId> ,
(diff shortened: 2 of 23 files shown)


brianjlowe (changes requested):
These changes look fine with the exception of the permissions. Deploying this with an existing VIVO database yielded a ROR ID with no display or edit permissions checked (see screenshot). Other identifiers default to everything checked except public (second screenshot). We may need @litvinovg 's advice here, as I might have expected the added annotations to migrate automatically. **Edit:** Actually, the other properties aren't quite right as they should be defaulting to public publishing even if they're not automatically displayed on the public profile page. So the last row of checkboxes should have all of them checked. We probably need a separate issue to fix the existing identifiers.

Found:
Image

Expected:
Image

litvinovg:
> These changes look fine with the exception of the permissions. Deploying this with an existing VIVO database yielded a ROR ID with no display or edit permissions checked (see screenshot). Other identifiers default to everything checked except public (second screenshot). We may need @litvinovg 's advice here, as I might have expected the added annotations to migrate automatically.

Permissions should be added to the respective permission sets. For [example](https://github.com/vivo-project/VIVO/blob/08d32c650a1fa8196e6352bd05a3d4ec8ca9b3a8/home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_admin_add_data_property.n3#L7)

chenejac:
@brianjlowe I have manually added permissions for ROR ID aligned with your second picture. However, we might discuss whether ROR ID should be displayed for public.

home/src/main/resources/rdf/accessControl/firsttime/allowed_entities_public_publish_data_property.n3
```diff
@@ -182,6 +182,7 @@
     vivo:rank ,
     vivo:reportId ,
     vivo:researchOverview ,
+    vivo:rodId ,
```
litvinovg:
```suggestion
    vivo:rorId ,
```

litvinovg:
Currently new rorId literals contain language tag, which is not what we need in this case.
The suggestion is to limit the range of vivo:rorId to xsd:string to avoid that.

litvinovg (approved):

brianjlowe (approved):
Confirmed that permissions are added on a new installation and that editing works as expected. (The new publish permission didn't appear when installing over an existing database, but that's clearly not related to this PR; it may suggest a possible separate issue with the update of the access control firsttime RDF.)

PR #4194, opened 2026-09-02
OAI PMH fixes

litvinovg:
Fixes for OAI PMH review comments
[Mapping implementation fixes ](https://github.com/litvinovg/rdf-to-openaire/commits/main/?since=2026-09-02&until=2026-09-04)
@chenejac

### DIFF
--- a/home/src/main/resources/rdf/display/firsttime/oai_pmh_distributors.n3
+++ b/home/src/main/resources/rdf/display/firsttime/oai_pmh_distributors.n3
@@ -460,7 +460,9 @@ WHERE {
 		CONSTRUCT {
 		  ?org ?orgP ?orgO .
 		  ?uri rdfs:comment "OpenAIRE OrgUnit" .
+		  ?org vivo:rorId ?rorId .
 		  ?greaterOrg obo:BFO_0000050 ?org .
+		  ?greaterOrg vivo:rorId ?greaterOrgId .
 		  ?org obo:ARG_2000028 ?vcardInd .
 		  ?vcardInd vcard:hasURL ?urlInd .
 		  ?urlInd vcard:url ?url .
@@ -478,6 +480,12 @@ WHERE {
 		  ?uri a foaf:Organization .
 		  OPTIONAL {
 		    ?greaterOrg obo:BFO_0000050 ?org .
+		    OPTIONAL {
+		      ?greaterOrg vivo:rorId ?greaterOrgId .
+		    }
+		  }
+		  OPTIONAL {
+		    ?org vivo:rorId ?rorId .
 		  }
 		  ?org a|rdfs:label|vitro:mostSpecificType|vivo:abbreviation ?orgO .
 		  ?org ?orgP ?orgO .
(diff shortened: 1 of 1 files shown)


home/src/main/resources/rdf/display/firsttime/oai_pmh_distributors.n3
```diff
@@ -207,6 +230,7 @@ WHERE {
 		  ?publication ?p ?o .
 		  ?publication vivo:relatedBy ?relBy .
 		  ?relBy a ?relType .
+		  ?relBy vivo:rank ?rank .
```
chenejac, line 233: Can this be also added for sparql queries for Patents and Products?

home/src/main/resources/rdf/display/firsttime/oai_pmh_distributors.n3
chenejac: vivo:Grant ?

chenejac (changes requested):
@litvinovg a couple of additional questions linked with xsl transformation

- <xsl:sort select="vivo:rank" order="ascending" /> Is this numerical or alphabetical order? I mean 10 won't be before 2?
- Type for Publication, Product, Patent, Funding should be mandatory in OpenAIRE Guidelines for CRIS manages (no minOccurs="0"). I am afraid exported records might be invalid if there is no Type. Can we include there some fallback (otherwise) with some default COAR resource type? For instance, it might be Other - https://vocabularies.coar-repositories.org/resource_types/c_1843/ for Publication, Product, Patent? For Funding, I am not sure what we can do. 
- Do we need `<xsl:if
				test="starts-with(res:id(vivo:dateTimePrecision),'http://vivoweb.org/ontology/core#yearMonthDay') and string-length(vivo:dateTime/text()) &gt; 9">
			` for registrationDate for Patent? It is defined as xs:date in OpenAIRE Guidelines

home/src/main/resources/rdf/display/firsttime/oai_pmh_distributors.n3
litvinovg: fixed

home/src/main/resources/rdf/display/firsttime/oai_pmh_distributors.n3
litvinovg: done

litvinovg:
> @litvinovg a couple of additional questions linked with xsl transformation
> 
>     * <xsl:sort select="vivo:rank" order="ascending" /> Is this numerical or alphabetical order? I mean 10 won't be before 2?
> 
Explicitly specified number sorting
>     * Type for Publication, Product, Patent, Funding should be mandatory in OpenAIRE Guidelines for CRIS manages (no minOccurs="0"). I am afraid exported records might be invalid if there is no Type. Can we include there some fallback (otherwise) with some default COAR resource type? For instance, it might be Other - https://vocabularies.coar-repositories.org/resource_types/c_1843/ for Publication, Product, Patent? For Funding, I am not sure what we can do.
Set fallbacks for publications and products.  No need for fallback for patent types. For funding both possible types have mapping, so fallback also not needed.
> 
>     * Do we need `<xsl:if test="starts-with(res:id(vivo:dateTimePrecision),'http://vivoweb.org/ontology/core#yearMonthDay') and string-length(vivo:dateTime/text()) &gt; 9"> ` for registrationDate for Patent? It is defined as xs:date in OpenAIRE Guidelines
Yes, to check that precision specifies year, month and day and actual data contains at least 10 characters, required for year, month and day.

chenejac (approved):

